PR #42, opened 2015-09-06
Bugs

marksreeves:
Updated womponent_themes JavaScript wc/ui/selectLoader.js to provide a
workaround for an intermittent bug in Safari 8.0.8 which, when
triggered, causes a SELECT (not multiple) to fail to update its
selectedIndex when an option is programatically selected. The
circumstances are difficult to reproduce and need some convoluted UI
conditions.

Updated XSLT for ui:dateField to remove the placeholder when
allowPartial but not required. The placeholder content was tied to the
default display fomat and the default display format causes an aural
output error in NVDA.


marksreeves:
Ignore this.

PR #43, opened 2015-09-07
Ported bug fixes

marksreeves:
Put in a workaround for a Safari 8.0.8 bug which resulted in a select
element not updating its exposed selected option when selectedIndex is
updated programatically under some very specific conditions. Removed
the date format placeholder from ui:dateField XSLT as it (a) causes
issues in NVDA and (b) does not really reflect the range of allowed
input options.


ricksbrown:
@marksreeves I merged this work back into `release1` for you. Let me know if there was anything else that had to go in that direction.

PR #57, opened 2015-09-11
Rename pom group id.

jonathanaustin:
Pom group rename as per discussions.

@ricksbrown PTAL


ricksbrown:
@jonathanaustin LGTM
This needs to be merged into release2, I'll handle that once you have merged into release1.

Issue #46, opened 2015-09-08
WebUtilities.updateBeanValue(WComponent) should ignore non-visible components


Joshua-Barclay:
The standard approach within WComponents is that invisible components (component.isVisible() == false) are completely ignored from all phases of processing (e.g. validation, prepare paint component).

WebUtilities.updateBeanValue(Wcomponent) does not ignore invisible components and will invoke the update bean value method on the component and any children.

An overloaded method should be created with an additional boolean parameter specifying whether to ignore invisible components or not.

The default single-arg method should ignore invisible components, as this is the standard within the framework.

jonathanaustin:
Yes, this behaviour has probably been in WComponents for a very long time. It is not consistent and should be fixed.

However, it will have a potential "knock on" affect to some applications if they were depending on this behaviour. For example, if applications were using a "card manager" to process a "wizard" like form, the invisible cards would have still been pushed to the beans. So with the fix, this will not occur by default and will fail silently. Applications will have to change their code to call the new method with "true" to process the invisible components.

We could consider having the fix as an "opt in" or at least issue a warning for the invisible components that are not processed.

ricksbrown:
Closed by PR #108

PR #66, opened 2015-09-13
WebUtilities.updateBeanValue() isVisible check

Joshua-Barclay:
Additional logic to handle whether or not to update bean values for
invisible components.


ricksbrown:
Awesome thanks for this @Joshua-Barclay. Sorry about the failing travis build, there seems to be a concurrency issue in our tests, I kicked it off twice more until it passed.

@jonathanaustin PTAL

Closes #46

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WebUtilities.java
```diff
@@ -525,9 +525,16 @@ public static UIContext findClosestContext(final String id)
      * and all its bean-bound children.
      *
      * @param component the component whose contents need to be copied to the bean.
+     * @param ignoreVisible - whether to ignore visible components.
      */
-    public static void updateBeanValue(final WComponent component)
+    public static void updateBeanValue(final WComponent component, final boolean ignoreVisible)
```
jonathanaustin: I think the "ignoreVisible" might be confusing. Is it meant to be ignoring "invisible"?

There is bit of a convention in WComponents of using "visibleOnly" as the flag. (e.g. the getComponentById method in this class, or look at TreeUtil). So would be good to switch the flag to "visibleOnly" and then pass "true" to only process the visible components and "false" to process all.

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/WebUtilities_Test.java
```diff
@@ -517,6 +518,85 @@ public void testGetParentNamingContext()
...
+        
+        WContainer invisibleContainer = new WContainer();
+        invisibleContainer.setVisible(false);
+        root.add(invisibleContainer);
+        WTextField childOfInivisbleContainerTextField = new WTextField();
+        childOfInivisbleContainerTextField.setBeanProperty(childOfInvisibleContainer);
+        childOfInivisbleContainerTextField.setData(childOfInvisibleContainer);
+        childOfInivisbleContainerTextField.setFlag(ComponentModel.USER_DATA_SET, true);
+        invisibleContainer.add(childOfInivisbleContainerTextField);
+
```
jonathanaustin: To save calling the USER_DATA_SET flag (as you should not have to call this), you need to "lock" the component tree so it treats the test like a "new user" session. When the tree is locked this flag will be set. It is not being set at the moment as WComponents is treating any values "Set" as the "defaults".

So after all the components have been setup, you can include this....

root.setLocked(true);
setActiveContext(createUIContext());

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/WebUtilities_Test.java
```diff
@@ -517,6 +518,85 @@ public void testGetParentNamingContext()
         Assert.assertEquals("Naming context for child3 should be context3", context3,
                             WebUtilities.getParentNamingContext(child3));
     }
+    
+    /**
+     * Test the function of updateBeanValue() for all variants (single argument, ignoreVisble [true|false]).
+     */
```
jonathanaustin: Could be a candidate for a parameterised test...

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WebUtilities.java
```diff
@@ -545,11 +552,21 @@ public static void updateBeanValue(final WComponent component)
...
-
+    
+        /**
+     * Updates the bean value with the current value of the component
+     * and all its bean-bound children. By default this method will process invisible containers and inputs.
+     *
+     * @param component the component whose contents need to be copied to the bean.
+     */
+    public static void updateBeanValue(final WComponent component)
+    {
```
jonathanaustin: If the flag changes to "visible only", then this will change to default "true". As this will have a possible impact to projects as they might have been getting some unexpected behaviour (i.e. the invisible components being updated). This will need to be identified as an impact in the change log.

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/WebUtilities_Test.java
```diff
@@ -517,6 +518,85 @@ public void testGetParentNamingContext()
         Assert.assertEquals("Naming context for child3 should be context3", context3,
                             WebUtilities.getParentNamingContext(child3));
     }
+
```
jonathanaustin: Test looks good. Very thorough!

Joshua-Barclay:
I've committed changes to resolve the issues you've raised.

I have avoided a parameterised test as I believe that will run the entire test class for each set of parameters. I do not think this method warrants a standalone test class, especially given how easy it is to test the variations (Boolean will only ever have 3 variants).

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WebUtilities.java
```diff
@@ -525,9 +525,16 @@ public static UIContext findClosestContext(final String id)
      * and all its bean-bound children.
      *
      * @param component the component whose contents need to be copied to the bean.
+     * @param visibleOnly - whether to include visible components only.
      */
-    public static void updateBeanValue(final WComponent component)
+    public static void updateBeanValue(final WComponent component, final boolean visibleOnly)
```
jonathanaustin: This looks good. Easier to understand.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WebUtilities.java
```diff
@@ -545,11 +552,21 @@ public static void updateBeanValue(final WComponent component)
...
+    
+        /**
+     * Updates the bean value with the current value of the component
+     * and all its bean-bound children. By default this method will process invisible containers and inputs.
+     *
+     * @param component the component whose contents need to be copied to the bean.
+     */
+    public static void updateBeanValue(final WComponent component)
+    {
+            updateBeanValue(component, false);
```
jonathanaustin: Did you mean this to be true? Did we want to make "true" the new default behaviour so only "visible" is processed? We will then have to explain this in the change log.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WebUtilities.java
```diff
@@ -525,9 +525,16 @@ public static UIContext findClosestContext(final String id)
      * and all its bean-bound children.
      *
      * @param component the component whose contents need to be copied to the bean.
+     * @param visibleOnly - whether to include visible components only.
      */
-    public static void updateBeanValue(final WComponent component)
+    public static void updateBeanValue(final WComponent component, final boolean visibleOnly)
     {
+        //Do not process if component is invisble and ignore visible is true. Will ignore entire branch from this point.
```
jonathanaustin: Style nit: space between // and start of comment.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WebUtilities.java
Joshua-Barclay: Based on the discussions we had I decided to leave it false to ensure no impact to existing applications.

Personally I'd prefer it to process visible components only, but I understand the need to reduce upgrade impacts.

jonathanaustin: I think if the change is documented in the change log as an API change, it is worth making it "true", as projects could be getting unexpected behaviour with the current code. They can always opt back into the old behaviour.

Joshua-Barclay: Yep agreed. I'll change the default to true (visible components only) and update the change log.

Joshua-Barclay:
@jonathanaustin I've updated the default to true and fixed the style issue. Please review.

jonathanaustin:
LGTM

@Joshua-Barclay Thanks. Awesome!

@ricksbrown Please merge once you have applied the format changes.

ricksbrown:
@Joshua-Barclay FYI because we have changed our code formatting guidelines dramatically since you opened this PR I'm going to pull your commits into another branch, apply the new formatting and then open a new PR with your work (it will still be attributed to you - the commits will still be yours with an additional reformatting commit from me).

PR #79, opened 2015-09-16
add acknowledgements

marksreeves:
add acknowledgements page to maven site


marksreeves:
@jonathanaustin When you have some time could you please go through the old Maven site and make a list of pages you think you want/need to keep. I am happy to move them to markdown but cannot determine which are absolutely necessary and which should be moved to the wiki or removed.

PR #82, opened 2015-09-17
Get rid of products directory and third party binaries

ricksbrown:
xslmin was the last one - now published to maven central


ricksbrown:
oops wrong base branch

PR #102, opened 2015-09-24
Release1: Fix Sass issue in menus

marksreeves:
WMenu Types BAR and FLYOUT with WSubMenu more than one level deep: the submenu UI artefacts are incorrectly rendered due to a CSS error. 

@ricksbrown I will apply a manual patch to Release2: no need to merge up.


ricksbrown:
Ok no problemo.

PR #105, opened 2015-09-26
Update ComponentModel API so that it manages its own internal List of children

javapod:
Currently ComponentModel allows its internal state to be directly modified externally. This PR proposes to modify the API so that ComponentModel fully encapsulates its internal list of children and not allow this internal list to be directly accessible or modifiable externally.

The API will be changed so that:
_addChild(WComponent child)_ is used to add children to the model. If the internal children list is null then create a new list and add the child. Return true if the child was added.

_removeChild(WComponent child)_ is used to remove a child from the model. If the children list is null, then return false. If the child was removed, and the children list is now empty, then deallocate the list to save memory. If the child was removed and there are other children present, then trim the list to save memory. Return true if the child was removed.

_hasChildren()_ is a convenience method that helps to hide the implementation of how the children are stored.

_getChildren()_ returns an unmodifiable List<WComponent> that prevents the internal list from being modified.

_setChildren(List<WComponent> children)_ creates a defensive copy of the list passed in to prevent the internal state from leaking out. If the parameter is null or an empty list then the internal children list is assigned to null, to save memory.

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
@@ -568,11 +568,10 @@ public abstract class AbstractWComponent implements WComponent {
 		// Prepare this component.
 		preparePaintComponent(request);
 
-		// Prepare its children.
-		final ArrayList<WComponent> children = (ArrayList<WComponent>) getComponentModel().
-				getChildren();
+		if (getComponentModel().hasChildren()) {
 
-		if (children != null) {
+			// Prepare its children.
+			final ArrayList<WComponent> children = (ArrayList<WComponent>) getComponentModel().getChildren();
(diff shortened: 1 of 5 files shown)


ricksbrown:
@javapod it certainly reads better this way.
@jonathanaustin PTAL

javapod:
Maven runs successfully when I run locally. Not sure why travis is failing.

jonathanaustin:
Looking at the Travis build log it shows a checkstyle error

```
/home/travis/build/BorderTech/wcomponents/wcomponents-
core/src/test/java/com/github/bordertech/wcomponents/
ComponentModel_Test.java:22:56: ',' is not followed by whitespace.
```
As we are still reviewing the checkstyles, please feel free to provide any feedback.

jonathanaustin:
I will take a look at the changes. Curious what made you look at the ComponentModel as there were no open issues.

PR #109, opened 2015-09-29
Update schema doc

marksreeves:
Updated schema documentation. Added docs for ui:figure (WFigure). Updated images. 

@ricksbrown optional: we could merge commit 1d1de7e into Release1 as its schemadoc is incorrect but this is pretty minor IMO and it is a reasonably big update with 271 changed files.


ricksbrown:
Absotively posilutely not worth it.

PR #119, opened 2015-10-06
Fix for issue #116 add getChildren() to Container API

### LINKED ISSUE #116: Add getChildren() to Container API
Adding this method to the Container API would allow simpler iteration using the enhanced for loop.

For example, in short it would turn this:

```
if (component instanceof Container) 
{
    Container container = (Container) component;
    int childCount = container.getChildCount();

    for (int i = 0; i < childCount; i++) 
    {
        WComponent child = container.getChildAt(i);
        // Do something with child....
    }
}
```

into this:

```
if (component instanceof Container) 
{
    for(WComponent child : ((Container)container).getChildren())
    {
        // Do something with child....
    }
}
```

The return type would be List<? extends WComponent>, and implementations should return an unmodifiable List. If there are no children then an empty list should be returned.

Importantly, the existing 3 methods on Container are all provided by List:

size() -> getChildCount()
get(int i) -> getChildAtIndex()
indexOf(Object o) -> getIndexOfChild(WComponent child)

If this is accepted then the existing methods on Container could/should be deprecated?

JohnMcGuinness:
fixes #116


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WebUtilities.java
```diff
@@ -502,9 +504,14 @@ public static void updateBeanValue(final WComponent component, final boolean vis
 		}
 
 		if (component instanceof Container) {
+
+			final List<WComponent> children = ((Container) component).getChildren();
+
+			Collections.reverse(children);
```
JohnMcGuinness: Invoking reverse on an unmodifiable List will probably throw an exception. Might need to add the children to a new List before reversing.

ricksbrown:
@JohnMcGuinness thanks for the PR.
@jonathanaustin PTAL

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
```diff
@@ -436,11 +437,11 @@ public final void serviceRequest(final Request request) {
 	 */
 	private static void collateVisible(final WComponent component, final List<WComponent> list) {
 		if (component.isVisible()) {
+
 			if (component instanceof Container) {
-				final int size = ((Container) component).getChildCount();
 
-				for (int i = 0; i < size; i++) {
-					collateVisible(((Container) component).getChildAt(i), list);
+				for (WComponent child : ((Container) component).getChildren()) {
```
jonathanaustin: The new getChildren() method is looking really good and makes the API easier to use.

As every component extends AbstractWComponent, we have to be careful of performance.

The new getChildren() method needs to be looked at as a "public" API - its job is to make writing code easier for application developers. So it does make sense it returns the list of children as an unmodifiable list to protect internal data from external code.

But when AbstractComponent needs to work internally with the list of children, it does not need to have the overhead of creating an unmodifiable list - we are not trying to make life easier for framework developers.

So for performance, it would be better that AbstractWComponent continue to use the list directly off the model.

Even though we don't get to use a nice "for each" pattern internally it is better for performance as we don't create the unmodifiable list which only exists to protect internal data from external code.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
```diff
@@ -1278,8 +1279,7 @@ public boolean isDefaultState() {
 	 * @return the number of child components currently contained within this component.
 	 */
 	int getChildCount() {
-		ComponentModel model = getComponentModel();
-		return (model.getChildren() == null ? 0 : model.getChildren().size());
+		return getChildren().size();
```
jonathanaustin: As above...again it would be better to just get the size off the model without creating another list.

If you can go through the PR and apply this principle, that would be great. "@" me when you have made the changes and I will take another look.

Thanks again for the PR as it will be a good API change for application developers.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
JohnMcGuinness: @jonathanaustin Thanks for the feedback. How about if getChildren returns an unmodifiable list as my PR uses but if I add another method getModifiableChildren() that passes through the List from the ComponentModel or an empty list if the Component Model's children is null. Collections.emptyList() returns a singleton instance so there is no overhead of constructing a new list.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WebUtilities.java
jonathanaustin: This is a good example of how the original way of using indexes to access the list is better for performance.

JohnMcGuinness:
@jonathanaustin I've made the changes from your feedback. PTAL

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WField.java
```diff
@@ -100,10 +100,10 @@ public String getId() {
 	}
 
 	/**
-	 * Finds the first component that can be labelled (if any).
+	 * Finds the first component that can be labeled (if any).
 	 *
 	 * @param component the root of the component hierarchy to check.
-	 * @return the first component that can be labelled, otherwise null.
+	 * @return the first component that can be labeled, otherwise null.
```
ricksbrown: We probably need a policy on US vs GB spelling...

That said I'll spell "color" with a superfluous "u" when the proverbial freezes over

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMenu.java
```diff
@@ -378,8 +378,8 @@ private void findSelections(final Request request, final Container component,
 			if (child instanceof WMenuItemGroup) {
 				WMenuItemGroup group = (WMenuItemGroup) child;
 
-				// Grouping doesn't affect selectability.
-				// Groups can not be nested, so just loop through the group's children.
+ 				// Grouping doesn't affect selectability.
+ 				// Groups can not be nested, so just loop through the group's children.
```
ricksbrown: Why the spaces before the indent?

wcomponents-examples/src/main/java/com/github/bordertech/wcomponents/examples/picker/SafetyContainer.java
```diff
@@ -152,7 +152,7 @@ protected void paintComponent(final RenderContext renderContext) {
 			} else {
 				try {
 					for (int i = 0; i < shim.getChildCount(); i++) {
-						shim.getChildAt(i).paint(renderContext);
+    						shim.getChildAt(i).paint(renderContext);
```
ricksbrown: Style nit: please use consistent indentation characters (i.e. tabs)
@jonathanaustin looks like your checkstyle rules are missing this scenario.

JohnMcGuinness:
@ricksbrown @jonathanaustin PTAL

jonathanaustin:
Looks good. Thanks!

jonathanaustin:
The pull request is looking good. Can you make sure you have merged the latest changes from release2 and then we will merge the pull request. Thanks.

JohnMcGuinness:
@jonathanaustin Merged with Release 2

PR #145, opened 2015-10-19
Add getUnmodifiableChildren to ComponentModel

### LINKED ISSUE #132: Add getUnmodifiableChildren() to the ComponentModel API
I realize I may be naively jumping in the deep end by proposing this, but I'll put my reasons never the less. I would appreciate feedback for or against my rationale.

Since Container has had the `getChildren()` method added it is an API that now has two faces. On the one hand are application developers, where `getChildren()` is favoured, because if its ease of use. And on the other hand are the remaining methods `getChildAt()`, `getChildCount()` and `getIndexOfChild()` that are favoured by the framework developers, but also may be used by application developers, because of their better performance.

I am proposing that `ComponentModel` has a new method `getUnmodifiableChildren()` added. Doing this would provide ease of use and performance for Application and Framework developers.

There are two possible approaches that I can see, in both cases a new field is added:

```
    private List<WComponent> unmodifiableChildren;
```

This field may or may not be final depending on the implementation chosen.

The first option has the field being declared final, the field is initialised with:

`private final List<WComponent> unmodifiableChildren = Collections.unmodifiableList(children);`

and the `getUnmodifiableChildren()` method would then look like this:

```
    protected List<WComponent> getUnmodifiableChildren() {
        return children == null ? Collections.<WComponent>emptyList() : unmodifiableChildren;
    }
```

I prefer this option because it doesn't require any object creation at runtime, therefore giving Application developers the performance that Framework developers get. It also gives Framework developers the ease of use the Application developers get. However, it does require the ComponentModel to use more memory to refer to the `unmodifiableChildren` field. Memory is cheap :), or at least last time I checked it was.

The second option has the field not being declared final, to enable dynamic allocation of the field. This option may then require changes to the `setChildren()` method to dynamically allocate and deallocate the unmodifiableChildren list. The `getUnmodifiableChildren()` would then look like this:

```
    protected List<WComponent> getUnmodifiableChildren() {
        if(children == null) {
            return Collections.<WComponent>emptyList();
        } else {
            if (unmodifiableChildren == null) {
                unmodifiableChildren = Collections.unmodifiableList(children);
            }
            return this.unmodifiableChildren;
        }
    }
```

and `setChildren()` would look like this:

```
    protected void setChildren(final List<WComponent> children) {
        this.children = children;
        if(children == null) {
            this.unmodifiableChildren = null;
        } else {
            if (unmodifiableChildren == null) {
                this.unmodifiableChildren = Collections.unmodifiableList(children);
            }
        }
    }
```

So what are the pros and cons?

The reasons in favour of adding `getUnmodifiableChildren()` include:
- The implementation of `AbstractWComponent::getChildren()` is trivial. It would be:

```
public List<WComponent> getChildren() {
    getComponentModel().getUnmodifiableChildren();
}
```

it simply passes through the `ComponentModel`'s unmodifiable children.
- It removes the need to continuously create unmodifiable lists as `getChildren()` in `AbstractWComponent` currently does.
- Because `getChildren()` passes through an unmodifiable view of the `ComponentModel`'s children the framework can use `Container::getChildren()` itself, assuming the implementation uses `AbstractWComponent::getChildren()`.
- Looking to the future when WComponents runs on JDK8, it will be trivial to convert the API to return a `Stream<WComponent>` which in turn will enable the framework to process children concurrently. Even better performance :).
- It provides Application developers and Framework developers the the same performance and ease of use.

The main reason against adding `getUnmodifiableChildren()` is the additional memory usage by `ComponentModel`, maybe that is enough to sink this change :(.

 @jonathanaustin @ricksbrown WDYT?

I think it is absolutely essential that `Container::getChildren()` is present going forward in JDK8+ because of the language changes available in JDK8.

This change would also need to be tested to measure the performance of any change. I will need some guidance on how to do that.

JohnMcGuinness:
Fixes #132


jonathanaustin:
Thanks John.

Could you also include a new Junit to make sure the list returned by the new method is still unmodifiable when a new Model is created for a user in a new context.

Also test for the second user that if a new "child" is added it is only available in the list to the second user.

marksreeves:
Hi @JohnMcGuinness how are things going with @jonathanaustin's request for JUnits? Is there anything we can help you with?

Issue #13, opened 2015-08-26
Fix maven archetype


ricksbrown:
It's been neglected for a loooooooooooong time

marksreeves:
Given that a working application must reference WComponents core and a theme perhaps the archetype is a separate project?

ricksbrown:
It should be updated to reference `wcomponents-bundle`.

jonathanaustin:
Fixed by #248.

PR #165, opened 2015-10-29
App archetype

marksreeves:
Updated the Maven archetype. It now creates an _almost_ working archetype.

Still to do:
* The WComponents version needs to be set manually in the generated
projects POM.
* The wcomponents-lde dependency in the generated POM is incompatible
with actually running the LDE because of the provided scope.
* The search results uses WDataTable and this should be updated to use
WTable.
* Fix war packaging because I can’t make it work with GlassFish, but I
did not try very hard.

Here are the steps required to get a generated app started:

1. Install wcomponents-app-archetype;
2. generate an archetype based application;
3. open the generated project;
4. edit the generated POM: update the WComponents version to
1.0.0-SNAPSHOT and (if using NetBeans andyou want to run the LDE)
delete the `provided` scope on the dependency for wcomponents-lde;
5. run the project (if using the LDE chose the PlainLauncherProxy).

Addresses #13.


marksreeves:
@jonathanaustin would you care to take a look at the archetype before we merge these changes? It is a _teensy_ bit dodgy...

marksreeves:
Withdrawn pending determination of the future of the archetype.

PR #175, opened 2015-11-04
Theme inheritance

### LINKED ISSUE #173: Theme inheritance restricted to same source tree
Building a theme which inherits from an existing implementation (not directly from wcomponents-theme) currently requires that the sub-theme and inherited theme are in the same source directory, for example:

```
src
 |- parent_theme
 |_ child_theme
```

This is overly restrictive as it prevents use of published themes (such as our own theme-inhouse project) as the basis of inheritance. I would argue that this is a bug rather than an enhancement because we have a sample theme which allows inheritance (through the existence of `sources.xml`) but which is not usable for this very process.

## Suggestion

We should allow inheritance from a path to another theme, including from the source zip file created during build. So given two independent paths:
```
/some/path/to/parent_theme
/some/other/path/to/child_theme
```

The child theme should be buildable having an `inherit.txt` containing either `/some/path/to/parent_theme` **or** `/some/path/to/a/previously/built/parent/theme/parent_theme-[VERSION]-sources.zip`.

marksreeves:
Updated sub-theme merging so that `inherit.txt` can contain one of:

1. a simple string without “/“ which is a theme in the current source
root (the currently supported scenario); or
2. the path to the root directory of the theme to be inherited; or
3. the path to a theme sources zip file.

Fixes #173

### DIFF
--- a/wcomponents-theme/build.xml
+++ b/wcomponents-theme/build.xml
@@ -233,15 +233,60 @@ chromedriver_cdnurl=${mirror.root}/chromedriver
 			<property name="impl.src.dir" location="@{rootDir}/@{implName}"/>
 			<var name="prototype.name" unset="true"/>
 			<loadfile property="prototype.name" srcFile="${impl.src.dir}/inherit.txt"/>
+
+			<var name="prototype.dir" unset="true"/>
+			<var name="merge.impl.dir" unset="true"/>
+			<var name="prototype.src.dir" unset="true"/>
+
+			<if>
+				<isset property="prototype.name"/>
+				<then>
+					<if>
+						<and>
+							<available file="${prototype.name}" type="file"/>
+							<matches string="${prototype.name}" pattern="\.zip$" casesensitive="false"/>
+						</and>
+						<then>
+							<echo>Found prototype zip file ${prototype.name}</echo>
+							<delete dir="${tmp.dir}/parent"/>
+							<!-- unzip the zip and set the path -->
+							<unzip src="${prototype.name}" dest="${tmp.dir}/parent" overwrite="true" failOnEmptyArchive="true">
+								<patternset>
+									<include name="*/**"/>
+								</patternset>
+							</unzip>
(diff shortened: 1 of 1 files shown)


marksreeves:
@ricksbrown Do you want to take a look at this before it is merged?

wcomponents-theme/build.xml
```diff
@@ -233,15 +233,59 @@ chromedriver_cdnurl=${mirror.root}/chromedriver
...
+							<then>
+								<echo>Found prototype directory named ${prototype.name}</echo>
+								<property name="prototype.dir" location="${prototype.name}"/>
+							</then>
+						</elseif>
+						<elseif>
+							<available file="${impl.src.rootdir}/${prototype.name}" type="dir"/>
+							<then>
+								<echo>Using simple fallback prototype ${impl.src.rootdir}/${prototype.name}</echo>
+								<property name="prototype.dir" location="${impl.src.rootdir}/${prototype.name}"/>
```
ricksbrown: Same deal here, may as well set `prototype.dir` at the top of this block

wcomponents-theme/build.xml
```diff
@@ -233,15 +233,59 @@ chromedriver_cdnurl=${mirror.root}/chromedriver
...
+						<then>
+							<echo>Found prototype zip file ${prototype.name}</echo>
+							<delete dir="${tmp.dir}/parent"/>
+							<!-- unzip the zip and set the path -->
+							<unzip src="${prototype.name}" dest="${tmp.dir}/parent" overwrite="true" failOnEmptyArchive="true">
+								<patternset>
+									<include name="*/**"/>
+								</patternset>
+							</unzip>
+							<property name="prototype.dir" location="${tmp.dir}/parent"/>
```
ricksbrown: Suggest moving this to the top of the `then` block and then replace usages of `${tmp.dir}/parent` with `${prototype.dir}`.

ricksbrown:
@marksreeves two optional suggestions, otherwise LGTM
In future we may want to consider allowing relative paths as well but for now it's not worth it.

PR #183, opened 2015-11-07
Add server side XSLT capability and fix core unit tests

### LINKED ISSUE #174: Enable server side XSLT by default
This is a small step on the road to replacing XSLT with Custom Elements (Web Components).
It also accepts the fact that browser vendors have lost interest in XML / XSLT. It is common for us to come across issues that only occur on HTML pages generated by XSLT, for example:

- Terrible memory leaks in IE8 [when page generated by XSLT](https://support.microsoft.com/en-us/kb/2032595)
- Firefox performance API completely broken in pages generated by XSLT [but XSLT gets low priority](https://bugzilla.mozilla.org/show_bug.cgi?id=804069#c8):

> Honestly, I don't know of anyone not working on something more important. XSLT is not seeing a whole lot of usage on the web.

- Original Android browser did not do XSLT at all.
- Chrome's implementation of Web Components [throws errors if the page is generated by XSLT](https://code.google.com/p/chromium/issues/detail?id=551089).
- Windows Phone 8.1 lost all ability to do client side XSLT #141 

Is XSLT a powerful transformation engine that ships with every major browser out of the box? Yes.
Has it seen massive uptake and support? No.

Ultimately I believe this is a case of the [principle of least power](http://www.w3.org/2001/tag/doc/leastPower) - people prefer JSON instead of XML not **despite** it being less powerful but **because** it is less powerful.

It's time to demote it, move it to the server and start replacing it with plain old HTML and HTML Custom Elements.

### LINKED ISSUE #182: wcomponents-core unit tests no longer run.
Surefire plugin is misconfigured.

When re-enabling them four tests in `WNumberField_Test` now fail :frowning:

ricksbrown:
Fixes #182 
Closes #174 

OK the XSLT is not enabled by default but it can be with a simple property switch.

@jonathanaustin Might be worth taking a look at `*.java` changes and maybe the `pom.xml` too.

Note: I have not injected the param `xslVendor` because it is scarcely used - we should be able to remove our need to sniff Internet Explorer in the transform.


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/servlet/ServletUtil.java
```diff
@@ -360,9 +361,9 @@ public static InterceptorComponent createInterceptorChain(final HttpServletReque
 				new UIContextDumpInterceptor(), new AjaxSetupInterceptor(),
 				new WWindowInterceptor(true), new WrongStepAjaxInterceptor(),
 				new ContextCleanupInterceptor(), new ValidateXMLInterceptor(),
```
jonathanaustin: Here and below...switch the order of ValidateXmlInterceptor TransformXMLInterceptor.

When the ValidateXMlInterceptor is enabled (should only be in dev) it will pick up any malformed XML and wrap it as an error and then pass it through the TransformXMLInterceptor.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/TransformXMLInterceptor.java
```diff
@@ -0,0 +1,160 @@
...
+ * It is enabled by setting the "bordertech.wcomponents.xslt.enabled" to true.
+ *
+ * @author Rick Brown
+ * @since 1.0.0
+ */
+public class TransformXMLInterceptor extends InterceptorComponent {
+	/**
+	 * The key used to look up the enable flag in the current {@link Config configuration}.
+	 */
+	public static final String PARAMETERS_KEY = "bordertech.wcomponents.xslt.enabled";
```
jonathanaustin: Make sure this new parameter is documented on the Wiki.

jonathanaustin:
LGTM once addressed comments.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/TransformXMLInterceptor.java
```diff
@@ -0,0 +1,160 @@
...
+	private static final Log LOG = LogFactory.getLog(TransformXMLInterceptor.class);
+
+	/**
+	 * Override paint to perform XML to HTML transformation.
+	 *
+	 * @param renderContext the renderContext to send the output to.
+	 */
+	@Override
+	public void paint(final RenderContext renderContext) {
+		boolean doTransform = Config.getInstance().getBoolean(PARAMETERS_KEY, false);
```
jonathanaustin: Lets turn it on by default.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/TransformXMLInterceptor.java
ricksbrown: Good call, will do

PR #206, opened 2015-11-16
Allow width and height to be String instead of int

### LINKED ISSUE #178: Allow width to be specified in units other than px
Components which support a WIDTH (such as WColumn) currently expect this width to be a pixel (WDialog) or percent (WColumn, WTableColumn, ColumnLayout) measure.

This should be any allowed unit in CSS. Therefore I suggest we change int width to String width.

marksreeves:
Update width of Column in WColumn, ColumnLayout and WTableColumn and
width of WDialog to be a string instead of an int. Deprecate
setWidth(int) in WColumn, WDialog and WTableColumn.

Update WDialog to allow height as a string (CSS dimension) instead of
an int. Deprecate setHeight(int).

@jonathanaustin Could you take a look through this please. It is to fix #178 and allows for setter backwards compatibility but not typed getter compatibility so I am not _completely_ content.

### Note on non-178 code in examples
There are a pile of tiny changes in JavaDoc in the wcomponents-examples project which I had to include to get a successful build after picking up your last changes. We need to sort out the JavaDoc because **random build failures** due to crusty JavaDoc in example code are sub-optimal. @ricksbrown could not replicate my issues. It seemed to be caused by the initial run of maven-javadoc-plugin after a version increment.


jonathanaustin:
Width and Height do need attention and allow more flexibility. There is also the "zero" height and width issue where a "zero" value cannot be set and "zero" defaults to the theme value.

Allowing just a "string" for the value and allowing the projects to set a "value" and "unit" (e.g. "20px") could be problematic. What if we provide some "type" safety.

We could introduce two new Interfaces "Widthable" and "Heightable". Maybe even a third interface "Sizeable" that combines both.

The interfaces could provide the following:-

```java
setWidth(BigDecimal value)
BigDecimal getWidth()

setWidthUnit(UnitEnum unit)
UnitEnum getWidthUnit()
```

Where UnitEnum is an enum of CSS Unit types ("ems", "%", "px" etc....). 

The same for "Heightable":-

```java
setHeight(BigDecimal value)
BigDecimal getHeight()

setHeightUnit(UnitEnum unit)
UnitEnum getHeightUnit()
```

Components that already have "setWidth(int)" would still be deprecated and call "setWidth(dec, "px")" under the hood. Same for "setHeight(int)".

The width unit could default to "ems" or whatever is appropriate.

The value should be validated to be null, zero or greater.

To address the "zero" value, if there is a "null" value, then the width is not rendered and the "theme" default is used. But a "zero" value is "rendered" as a zero.

Hope this makes sense. Please let me know what you think.

@marksreeves @ricksbrown

marksreeves:
I completely understand your points @jonathanaustin but have to disagree about dangers. Sure, if someone sets a width to `"Go that local sporting team"` then it will have undefined consequences (it will actually have no consequences, it will be as if it was never set but that is beside the point): that is not really _our_ problem.

If we are making a web framework how about we embrace the webbiness available to us. These dimensions represent direct access to CSS dimension properties in a **web** framework. It makes a little bit of sense IMO to let the app define the widths in these dimensions.

This may be controversial but I think we should:
1. care a little less about what applications do with the framework; and
2. assume people building web UIs can follow a web UI spec and that someone spec'ing a web UI will use common web dimensions.

@Joshua-Barclay you're more at the coal-face in this: what do you think?

marksreeves:
Oh, one more thing @jonathanaustin I am **very** happy to look at widthable and heightable interfaces - I had to really go hunting for the various classes which managed their own widths.

marksreeves:
As well as relative or absolute lengths or percentage width and height may both be calculated (using `calc()`) or use a keyword.

Keywords for width dimension:

* auto;
* available;
* fit-content;
* max-content;
* min-content.

Keywords for height dimension are as per width plus **complex**.

Keywords which can extend a length or percentage dimension for width or height:

* border-box;
* content box.

From [the W3C spec](https://drafts.csswg.org/css-box-3/#the-width-and-height-properties):

* width:
  
  ```
[<length> | <percentage>] && [border-box | content-box]? | available | min-content | max-content | fit-content | auto
  ```
* height:
  
  ```
[<length> | <percentage>] && [border-box | content-box]? | available | min-content | max-content | fit-content | complex | auto
  ```

These may make it a little bit hard to use an int and enum model.

marksreeves:
removed: alternate process determined.

Issue #247, opened 2015-12-05
wcomponents-app-archetype does not seem to work


JohnMcGuinness:
I built the wcomponents-app-archetype project (version 1.0.2-SNAPSHOT) and installed it locally.

I had a few issues when trying to use the archetype to create a new project, I've listed them below.

* When initially building the new project based on the archetype I received the error *'dependencies.dependency.version' for junit:junit:jar must be a valid version but is '${wcomponents.junit.version}'.*

 * **Solution:** Add the property  `<wcomponents.junit.version>4.8.2</wcomponents.junit.version>`

* I then received this compilation error: 
```
Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.5.1:compile (default-compile) on project Ache1: Compilation failure: Compilation failure:
com/github/johnmcguinness/ache1/servlet/MyAppServlet.java:[18,7] error: cannot access HttpServlet
class file for javax.servlet.http.HttpServlet not found
com/github/johnmcguinness/ache1/servlet/MyAppServlet.java:[32,4] error: method does not override or implement a method from a supertype
```

 * **Solution:** Added the following dependency: 
```
<dependency>
    <groupId>javax</groupId>
    <artifactId>javaee-web-api</artifactId>
    <version>7.0</version>
    <scope>provided</scope>
</dependency>
```

The project now successfully builds.

* I then tried to deploy to a GlassFish instance and received the following runtime exceptions when loading the app in a browser:
```
Warning:   StandardWrapperValve[myAppServlet]: Allocate exception for servlet myAppServlet
java.lang.ClassNotFoundException: org.apache.commons.logging.LogFactory
``` 
and 
```
Warning:   StandardWrapperValve[myAppServlet]: Allocate exception for servlet myAppServlet
java.lang.ClassNotFoundException: org.apache.commons.configuration.Configuration
```

 * **Solution:** Added the following dependencies: 
```
<dependency>
    <groupId>commons-logging</groupId>
    <artifactId>commons-logging</artifactId>
    <version>1.1.1</version>
</dependency>
``` 
and

```
<dependency>
    <groupId>commons-configuration</groupId>
    <artifactId>commons-configuration</artifactId>
    <version>1.6</version>
</dependency>
```

* I then received this runtime exception: `Info:   Service called on com.github.johnmcguinness.ache1.servlet.MyAppServlet@409b5746... from /Ache1/app
Severe:   Caught exception during action phase.
[...]

PR #276, opened 2015-12-17
Update archetype

JohnMcGuinness:
This should fix theme issue for #247


marksreeves:
Fixes #247

jonathanaustin:
@JohnMcGuinness Thanks for the change. I had bit more of a think about it....the wildcard "/app/*" could be just defined as one servlet mapping entry (instead of two "/app" and "/app/*"). Just a bit less noise in the web.xml and you could also adjust the comment in ThemeServlet. 

What do you think?

Unfortunately, to make the change you will need to do a new pull request.

JohnMcGuinness:
@jonathanaustin Tried just having the /app/ mapping, but the ui renders with no theme

PR #283, opened 2015-12-22
HTML Renderers

marksreeves:
WHorizontalRule xml was more verbose than the HTML equivalent. This
makes it an excellent candidate for testing render to HTML rather than
render to XML.

* remove the schema and XSLT for ui:hr
* update WHorizontalRuleRenderer to output `<hr` rather than `<ui:hr`
* update WHorizontalRuleRenderer_Test to test for html element rather
than ui:hr element


marksreeves:
@jonathanaustin Would you care to cast your eyes over this PoC for HTML renderers?

jonathanaustin:
LGTM

Issue #280, opened 2015-12-21
Allow WMessageBox (et al) title text to be set in application


marksreeves:
WMessageBox, WMessages, WValidationErrors each produce a UI artefact consisting of a titled section. These sections each have a title set by the theme and determined by the message severity, eg "Information" or "Warning".

The UI artefacts rendered by WMessageBox and WValidationErrors should allow the consuming application to set the text of the title. This has been requested by an existing consumer with a genuine use case and should be delivered before end Q1 2016.

It is acceptable that only WMessageBox have an over-rideable title if it is not able to be easily determined in WValidationErrors or WMessages.

The titles in wcomponents-theme are:

* ERROR (and WValidationErrors): "An error has occurred"
* WARNING: "Warning"
* INFO: "Information"
* SUCCESS: "Success"

Each also has an icon related to the message level. This may remain determined by the theme and therefore it is not necessary (at this stage) for the components to have a WDecoratedLabel (cf WSection, WCollapsible which are titled containers but of a much more generic nature).

marksreeves:
Stage one (as covered by feature request) is only for WMessageBox (note WMessages can access the WMessageBox to set/get the title). 

Stage two will be for extending this to WValidationErrors.

PR #287, opened 2015-12-22
WMessageBox title

marksreeves:
Added a String title member to WMessageBox’s model and accessors to WMessageBox. Updated renderer to output title if set. Updated unit tests to test new property, accessors and renderer.

Addresses #280.

@jonathanaustin would you care to take a look?


jonathanaustin:
LGTM.

You can use the new param on method [AbstractWComponentTestCase#assertAccessorsCorrect](https://github.com/BorderTech/wcomponents/blob/release2/wcomponents-core/src/test/java/com/github/bordertech/wcomponents/AbstractWComponentTestCase.java#L102)

For example....

```java
@Test
public void testMessageAccessors() {
 assertAccessorsCorrect(new WMenuItem(""), "message", null, "A", "B", new Serializable[]{});
}
```

Issue #262, opened 2015-12-16
WMenu API add getItems


marksreeves:
Offline enhancement request:

* add a method getItems to WMenu with the option to recurse through sub-menus.

**Priority**: _low_ (no specific use case with request)

jonathanaustin:
WMenu API allows for getItems(boolean recursive).

Introduced new interfaces MenuItem, MenuItemGroup, MenuItemSelectable, MenuContainer and MenuSelectContainer to support and enhance the menu components.

Issue #250, opened 2015-12-08
Improve loading a WSubMenu via AJAX


jonathanaustin:
At the moment, when a WSubmenu is loaded via "AJAX" it requires all its child menu items to already be added. 

It would be more flexible if a WSubmenu was not required to have its "children" added until the submenu is opened.

marksreeves:
We have promised this to an existing consumer and it is required in a release before the end of 2015.

jonathanaustin:
Fixed loading WSubMenus via AJAX.

Items can now be dynamically added when the WSubMenu is opened. 

Menu components now hold their child items in a Container and can be accessed via the getMenuItems() method. This allows the AJAX to target the container and the child menu items added dynamically. 

Refer to #262.

Issue #188, opened 2015-11-09
com.github.bordertech.wcomponents.AbstractWComponentTestCase.assertAccessorsCorrect does not work with setter Serializable args


marksreeves:
1. Create a JUnit to test accessors for which the setter has a Serializable argument;
2. use the helper `com.github.bordertech.wcomponents.AbstractWComponentTestCase.assertAccessorsCorrect` to test the accessors;
3. the JUnit fails unable to find setter;

Example
``` java
// getter
/**
 * {@inheritDoc}
 */
@Override
public String getHtmlClass() {
    ComponentModel model = getComponentModel();
    return I18nUtilities.format(null, model.getHtmlClass());
}

// setter
/**
 * {@inheritDoc}
 */
@Override
public void setHtmlClass(final String text, final Serializable... args) {
    ComponentModel model = getOrCreateComponentModel();
    model.setHtmlClass(text, args);
}

// JUnit
@Test
public void testHtmlClassAccessors() {
    assertAccessorsCorrect(new SimpleComponent(), "htmlClass", null, "foo", "bar");
}
```

Output
```
Caused by: java.lang.NoSuchMethodException: Property 'htmlClass' has no setter method in 
class ...
```

jonathanaustin:
AbstractWComponentTestCase.assertAccessorsCorrect now works with variable arguments on the setter methods.

If a setter method on the component has a variable argument, then an array of the variable argument type needs to be provided via the setterArgs parameter on the assertAccessorsCorrect method.

For example, if the variable argument is a type "Serializable" then the setterArgs parameter would be set as "new Serializable[]{}". Values can also be passed in via the array.

For example....

```java
@Test
public void testMessageAccessors() {
 assertAccessorsCorrect(new WMenuItem(""), "message", null, "A", "B", new Serializable[]{});
}
```

PR #290, opened 2015-12-23
Submenu ajax

jonathanaustin:
* WMenu API allows for getItems(boolean recursive). Introduced new interfaces MenuItem, MenuItemGroup, MenuItemSelectable, MenuContainer and MenuSelectContainer.  (#262)

* Fixed loading WSubMenus via AJAX. Items can now be dynamically added when the WSubMenu is opened. Menu components now hold their child items in a Container and can be accessed via the getMenuItems() method. This allows the AJAX to target the container and the child menu items added dynamically. (#250)

* AbstractWComponentTestCase now works with variable arguments on setter methods. If a setter method on the component has a variable argument, then an array of the variable argument type needs to be provided via the setterArgs parameter on the assertAccessorsCorrect method.
For example, if the variable argument is a type "Serializable" then the setterArgs parameter would be set as "new Serializable[]{}". Values can also be passed in via the array. (#188)


jonathanaustin:
* WMenu API allows for getItems(boolean recursive). Introduced new interfaces MenuItem, MenuItemGroup, MenuItemSelectable, MenuContainer and MenuSelectContainer.  (#262)

* Fixed loading WSubMenus via AJAX. Items can now be dynamically added when the WSubMenu is opened. Menu components now hold their child items in a Container and can be accessed via the getMenuItems() method. This allows the AJAX to target the container and the child menu items added dynamically. (#250)

* AbstractWComponentTestCase now works with variable arguments on setter methods. If a setter method on the component has a variable argument, then an array of the variable argument type needs to be provided via the setterArgs parameter on the assertAccessorsCorrect method.
For example, if the variable argument is a type "Serializable" then the setterArgs parameter would be set as "new Serializable[]{}". Values can also be passed in via the array. (#188)

PR #298, opened 2015-12-31
Pagination position

marksreeves:
Added a mechanism to allow applications to determine where to place pagination controls relative to a table.

Fixes 297


marksreeves:
@jonathanaustin Could you please take a look and I am done for the day so **happy new year**

Issue #299, opened 2016-01-01
Unify various Alignment enums into a single enum


JohnMcGuinness:
Currently there are numerous enums called `Alignment`. These are `ColumnLayout.Alignment`, `FlowLayout.Alignment`, `ListLayout.Alignment`, `WColumn.Alignment` and `WTableColumn.Alignment`.

All of these have the values `Alignment.LEFT`, `Alignment.CENTER` and `Alignment.RIGHT`. `FlowLayout.Alignment` also has the value `Alignment.VERTICAL`.

I would like to propose unifying all of these enums into a single enum `com.github.bordertech.wcomponents.layout.Alignment` with the values `LEFT`, `CENTER` and `RIGHT`. Also, to replace the `VERTICAL` value in `FlowLayout.Alignment`, I propose to add the `com.github.bordertech.wcomponents.layout.Orientation` enum with values of `HORIZONTAL` and `VERTICAL`.

A `FlowLayout` would by default have orientation of `Orientation.HORIZONTAL`, if the flow should be vertical then the developer could set the orientation:

```
myFlowLayout.setOrientation(Orientation.VERTICAL);
```

The vertically oriented `FlowLayout` could then still be aligned with an `Alignment` value.

WDYT @ricksbrown, @jonathanaustin, @marksreeves

marksreeves:
I think we should keep FlowLayout out of this. It has an alignment enum but it is not the same kind of thing as the other alignments (not just in the inclusion of `VERTICAL`). Column alignments are cosmetic and simply provide a general left-center-right alignment of the content in a given column (ColumnLayout, WColumn and WTableColumn). 

The Alignment enum in FlowLayout is structural and indicates the relationship of the layout's contained components to each other as well as their spatial arrangement on screen. It is a subtle but real difference which makes FlowLayout.Alignment fail the _is a_ test relative to the column alignments.

IMO this member should not be `Alignment` at all but that comes down to the tricky business of naming API constructs in an unambiguous and not-too-silly way (FlowLayout.Layout could be considered silly). Something like FlowLayout.Direction does not really indicate the nature of `CENTER` or `RIGHT` when it comes to a FlowLayout as they are still (default) left to right; FlowLayout.Arrangement is a bit florid (see what I did there?) but could be a bit more meaningful.

Maybe @jonathanaustin has something more sensible to add?

jonathanaustin:
[...]

PR #315, opened 2016-01-10
Unify alignment

JohnMcGuinness:
This mostly addresses the #299 with the exception of updating WFlowLayout. @marksreeves are you able to provide some advice about adding Orientation and updating Alignment for WFlowLayout.

PTAL @jonathanaustin, if its ok with you i would like to have this merged and address WFlowLayout later after Mark provides advice.


marksreeves:
I have no issues with melding Aligmnent for all of the "column" components but we need more discussion about FlowLayout. 

I think the notion of separating Orientation from Alignment is worthy in that it gives us six arrangements of components rather than the current four. The additional arrangements are: `VERTICAL` + `CENTER` and `VERTICAL` + `RIGHT`. I have seen a few UIs where `VERTICAL` + `CENTER` has been used with two nested FlowLayouts so this could be advantageous even though those particular UIs were rather poor examples as the centred content caused usability issues (and were rather ugly but that is, of course, just my personal taste).

Even if we decide to go down this route though this is a non-backwards-compatible API change so cannot be merged until after we release 1.0.3.

@jonathanaustin @ricksbrown do we have any idea when we can build 1.0.3? We promised it early this month (Jan 2016).

marksreeves:
@JohnMcGuinness Could you also have a think about an upgrade tool: if we make this API change it is going to have a massive impact on all consumers.

jonathanaustin:
Have added comments to #299.

JohnMcGuinness:
Just want to check what the concenus is before I update this PR.

Based on discussion in this PR and #299 this is what I propose to do:

* Rename `Alignment` to `CellAlignment` and include all 9 values.
* Remove the `Orientation` enum
* Add a new API that uses the `CellAlignment` values, and deprecate the 'old' `Alignment` enums and related methods.
* Leave FlowLayout as is, for now.

WDYT @jonathanaustin @marksreeves @ricksbrown 

@jonathanaustin If you would like I can have a go to create `HorizontalLayout` as in #329.

JohnMcGuinness:
@jonathanaustin PTAL I'm pretty sure this is ready to merge, based on various comments in #299 and this PR.

JohnMcGuinness:
@jonathanaustin any chance you look at this soon?

JohnMcGuinness:
@jonathanaustin any idea if/when this will be merged?

PR #318, opened 2016-01-13
Optional column width

### LINKED ISSUE #207: Column width should be optional
The width of a WColumn and the columns of a ColumnLayout should not be mandatory. This is a natural consequence of #172 and #178. 

If the widths are not defined then a theme may have defaults which may or may not be sensible but more importantly the application can use CSS to determine column layout and order of collapse for various viewports based on setting classNames on containers.

This will result in less inline CSS and significant improvements in support for **responsive UIs**. There will also be a reduction in payload size, especially when using server transforms.

marksreeves:
Updated WColumn and ColumnLayout to allow 0 width columns. This will then allow applications to use external CSS to style the column widths without having to resort to `!important` to override inline CSS.

Added examples to ColumnLayoutExample and WRowExample which include undefined (0) column width and 0 column width with application-level CSS using `set/getHtmlClass`. The latter examples include some rudimentary “responsive design” to show how it could be approached.

Fixes #207.


marksreeves:
@jonathanaustin want to take a look at this? It is backwards compatible at the moment using int 0 as the null/auto value.

jonathanaustin:
LGTM.

Using zero as undefined is OK and allows for backward compatability.

Might be good to add your explanation about the zero width to the xsd....

> zero width is a valid but meaningless width. So from a CSS point of view if I set soething to width:0 it
> would be honoured but without overflow control it would simply overflow. With columns width 0 is
> more problematic and would result in a broken UI so an optional attribute which is only output if > 0
> serves the same purpose without breaking any current implementations

Issue #151, opened 2015-10-21
Image manipulation


ricksbrown:
Add functionality to allow the following:

1. Image manipulation as part of an image file upload.
2. Image manipulation on an existing (already stored) image.

Image manipulation should include:
- crop
- align (move, perhaps relative to an overlay)
- zoom
- rotate

jepistola:
Would be good to have the functionality to rotate (eg. 45-90-180 degrees) an image inside a view panel and be able to zoom in / out the image in varying degrees.

ricksbrown:
@jepistola Yes I envisage a slider for rotation but with some presets (probably radio buttons) at logical rotations like the ones you mention.

Probably a similar case with zoom.

marksreeves:
## Suggestions for interactions and a11y helpers

* Cropping, rotating and re-sizing by filling in context-aware number fields (as per Photoshop, [Inkscape](https://inkscape.org/en/) etc), for example: 
    * top (crop);
    * left (crop);
    * width (crop, re-size);
    * height (crop, re-size);
        * example for crop (taken from Inkscape screen capture): <img width="289" alt="crop using input fields" src="https://cloud.githubusercontent.com/assets/11661938/10718683/c6cddea8-7bcc-11e5-9da4-a23c47709fd6.png" style="border:1px solid #999"> though I would use 'top' and 'left' rather than 'x' and 'y' even though this would cause extra i18n requirements and always enforce px as unit removing the need for the unit drop-down;
    * angle (rotate) - for example: <img width="87" alt="rotate_angle" src="https://cloud.githubusercontent.com/assets/11661938/10718834/7ed21738-7bcf-11e5-9485-6ea9c993bd85.png">;
    * center (rotate) - may be a bit much for general use.
* Context aware keyboard shortcuts, for example:
    * R/L rotate right (clockwise)/left (anticlockwise) 90 degrees;
    * Arrow keys for moving image within a crop box;
    * +/- keys for zoom.
* Toolbar (&/or menu) as part of the component which is configurable using a WMenu even if this is as restricted as allowing existing items to have their visibility or disabled state toggled. IMO using a property based build or run time configuration (as we do with the [tinyMCE](http://www.tinymce.com/) rich text add-on to WTextField) may leave the tool a bit weak compared with user context aware configuration.
* Integrated help.

## Some other options and possible requirements

1. Java API should have some simple setters to turn on/off basic sets of functionality, for example:
    * allow rotate - default true;
[...]

PR #319, opened 2016-01-13
Imageedit Experimental

ricksbrown:
This PR contains the following:

1. An experimental and unfinished solution to #151 Barring regressions there should be no impact as the imageedit Java API is entirely backwards compatible and the new JS code is not loaded unless it is explicitly needed.

2. A first baby-step to using a real templating solution to replace sprintf which was completely untenable as a templating tool.

I am primarily merging now because I am about to go on leave for the rest of January.


jonathanaustin:
Looks good so far.....

Might be good to try and keep WImage clean.....

How about extending WImage (eg WEditableImage) and use this to experiment with....

Issue #172, opened 2015-11-03
Simplify app-level front end component access


marksreeves:
At the moment it is a bit of a palaver to add CSS or JS customisation of individual instances of components (or groups of components) at the app level.

This allows an enterprise to tightly control appearance and function. It does, however, make it difficult to have exceptions or for individual applications which are slightly outside the regular corporate look & feel to implement their customisations. Such an app basically needs a custom theme or some rather labyrinthine Java workarounds.

I propose we make the process of front end customisation easier but in a way which allows an enterprise theme to disallow - by adding XML attribute(s).

In the first instance how about we add an XML attribute which allows a String HTML class attribute to be added to any UI component such as outlined in the following?

1. add an optional attribute (such as `@class`) to the schema for all UI components which output an element;
2. set/getHTMLClassName - maybe on AbstractWComponent - the HTML class name would be a String so if an application wanted to add multiple classNames they can simply by using a single space separated string as for HTML;
3. update renderers to output the HTML class name as an optional attribute; and
4. update default XSLT to output the string in HTML class attributes for each component.

Making this an XML attribute would allow a particularly tightly controlled theme to simply not output the XML `@class` attribute in the XSLT.

@jonathanaustin, @Joshua-Barclay, @JohnMcGuinness all comments welcome!

marksreeves:
I am re-opening this issue as I have been lead to believe the Java API for adding application-level CSS & JavaScript is too opaque.

To add a CSS file one has to use `Headers.addHeadline` or `Headers.addUniqueHeadline`. This has a few issues:

1. it is poorly documented;
2. it is not intrinsically sensible - what is a 'headline' and why would one want one?;
3. it is not exposed on any WComponent (WApplication _may_ be suitable);
4. it seems to conflate several concepts (or the JavaDoc is misleading - see 1) in that it uses the language of headers which imply HTTP headers but applies lines to the HTML head element but the JavaDoc states it can also apply attributes directly to the HTML element but this is not exposed in ui:root and therefore is not part of the XSLT.

Taking our universal 80% rule any user of WComponents should be able to add (preferably to a WApplication) a CSS link or a JavaScript src and have the framework deal with it adequately - everything else can be passed to custom templates.

@jonathanaustin Maybe you can clarify some of these for me?

marksreeves:
Re-open.

JohnMcGuinness:
[...]

PR #327, opened 2016-01-19
XSLT element class rationalisation

marksreeves:
Moved around the determination of className on the root element of most components to use a common helper template. This removes a lot of inconsistency in how the element local name and value of any `@class` attributes were applied.


marksreeves:
The point of this was to clean up some inconsistencies which were interfering with attempts to get #172 pushed along a bit more.

PR #334, opened 2016-01-21
Easier js css

jonathanaustin:
The new way of loading the javascript seems to run the script “too late” compared to actually putting it in the head line.

For example, the syntax highlighting in the examples does not work on the first page load, but if it is refreshed with AJAX it works.

Removed the “defer” when loading the js and it works.


jonathanaustin:
PTAL @marksreeves

marksreeves:
Go ahead and merge; I thought that might be an issue. The change is fine if that is the desired behaviour. The "working too late" is possibly a flaw in the primitive syntax highlighter but that is another issue.

jonathanaustin:
Thanks Mark. Will merge.

PR #348, opened 2016-01-27
257 row selection

### LINKED ISSUE #257: WTable row selection mechanism enhancement
We have received a request to add the following opt-in behaviour to row selection in WTable with sub-rows:

1. a WTable has sub-rows and multiple selection
2. any parent row is deemed to be a WSelectToggle analogue for its descendant rows if those rows are selectable
3. a parent row which is selectable is a simple selectable row if none of its sub-rows are selectable.

marksreeves:
Updated WTable so that when a table has multiple row selection and row expansion then “parent” rows may optionally act as “selectToggle” type controls for the child rows. Fixes #257.


marksreeves:
@jonathanaustin want to take a look at this? It is pretty straight-forward in the Java space but was a PITA in JavaScript. @ricksbrown we may want to revisit the whole selectToggle module at some point as it is really complex - it smells **too** complex.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTable.java
```diff
@@ -511,6 +511,25 @@ public void setSeparatorType(final SeparatorType separatorType) {
 	}
 
 	/**
+	 * Sets whether the de/selection of a row with sub rows should de/select the sub rows. This is a client-side only
+	 * feature.
+	 *
+	 * @param toggleSubRowSelection true to turn on this feature.
+	 */
+	public void setToggleSubRowSelection(boolean toggleSubRowSelection) {
```
jonathanaustin: Good practice to make this "final boolean".

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTable.java
```diff
@@ -511,6 +511,25 @@ public void setSeparatorType(final SeparatorType separatorType) {
...
+		getOrCreateComponentModel().toggleSubRowSelection = toggleSubRowSelection;
+	}
+
+	/**
+	 * Indicates whether de/selecting a row with sub row(s) will de/select the sub row(s).
+	 *
+	 * @return true if this feature is enabled.
+	 */
+	public boolean getToggleSubRowSelection() {
```
jonathanaustin: As it is boolean, the naming convention should be "isToggleSubRowSelection" on the getter.

PR #361, opened 2016-01-29
Table row selection sass cleanup

marksreeves:
a11y changes to menu bars caused the static menus in the tables used to
toggle sub-row selection to fail because I forgot to update them.

I also split the vertical and horizontal cell padding Sass variables to
fix an override issue in several sub-themes which was exposed by the CSS
changes needed to support the menu in the state display cell.

Finally, I found another condition of failure in selectToggle in which
a sub-row which also had sub-rows would fail to select its subrows when
a super-row controller was activated if said sub-row was already
selected.


marksreeves:
includes a patch for menus to alleviate the issue in #362 but does not remove that issue.

PR #382, opened 2016-02-09
306 WFilterControl

### LINKED ISSUE #306: Remove WDataTable FilterControl
`WDataTable` has been deprecated for more releases than we care to think about but is currently being kept as a **kindness** to a single consumer.

`WFilterControl` can _only_ act on `WDataTable` (not `WTable`) and is a poor implementation of a filter. We have provided examples of column based filtering and there are other AJAX powered filtering which is far more capable. The consumer for whom we are keeping WDataTable alive does not use `WFilterControl`.

We should remove `WFilterControl` if for no other reason than the performance gains in transforming tables and a significant reduction in XSLT complexity resulting in easier maintenance. However, since it applies only to a deprecated component and is not being used I see no reason to keep it.

marksreeves:
Fix #306.

**WARNING** This is a non-backwards compatible API change. There is no direct replacement for `WFilterControl`. Filtering should be done via AJAX. There is a simple example of this in wcomponent-examples.


marksreeves:
@jonathanaustin Thought I should put this by you. I quietly removed client support for access to this control from consumer themes last year and removed client side support from the wcomponents-theme in release 1.0.3.

I am also happy to remove the parts of `WDataTable` which refer to row filter values if you prefer. These do not directly interact with `WFilterControl`, it was just expected that the same string value[s]. They are now useless but do not impact the way the XSLT processing did.

I dropped some more explanatory info into the issue #306.

jonathanaustin:
LGTM.

As WDataTable has been deprecated for so long....what do you think of removing it with WFilterControl? 

The only thing is making sure WTable supports column row header that we can raise as a seperate issue and tag it for this release.

Trentan:
+1 WTable to support column row headers

PR #405, opened 2016-02-22
Validation and plugins

marksreeves:
* Moved validation plugin to wc/ui/validation.
* Added requirejs i18n module and used this to implement JS-only i18n in validation plugin.
* Added client side validation option to validation example.
* Removed build of “plugins” - this was always the wrong approach.


marksreeves:
* Updated the i18n example to have opt-in client side validation so we can see the effect of locale on client validation.

PR #502, opened 2016-04-05
switched sass compiler to wcomponents-sass-compiler

ricksbrown:
should dramatically improve build speed...


ricksbrown:
Elapsed times for theme rebuild (with tests skipped) on Windows...

Before this PR (using Vaadin Sass Compiler via the Ant `java` task):

* 5:14 min (4:04 min on sass compilation)

After this PR (using WComponents Sass Compiler):

* 1:14 min (3.5 sec on sass compilation)

PR #575, opened 2016-04-28
Responsive layouts/hgap/vgap/Sass

### LINKED ISSUE #571: Responsive GridLayout
Remove inline CSS from GridLayout to make responsive UIs easier to build.

### LINKED ISSUE #574: Suggestion's text input arrow click does not work
Given any WTextField etc which is linked with WSuggestions clicking on the "down arrow" is expected to open the suggestion list but it does not.

marksreeves:
It would be close to impossible to retain a backwards-compatible
implementation of GridLayout and make it abstractly responsive. These
changes remove a pile on inline CSS for most common GridLayout
implementations to make it possible for applications to attempt
responsive grids.

A corollary of this (in order to maintain consistency) is that hgap and
vgap have been changed **everywhere** so that inconsistent,
unresponsive and potentially inaccessible pixel based spacing has been
replaced by class-invoked CSS rem spacing in all cases and pixel to
class conversion is determined in the theme. The XSLT provides for ’n’
classes (default 4 for each) and the Sass provides the relevant gaps.
All of this is rather light weight so should be reasonably easy to
override both in a theme and in an application.

Most of the changes in this commit are a flow-on effect of this
rationalisation as it required re-implementing all gaps.

Fixes #571.

Changed the iconography of combo box fields to a background image
instead of a pseudo-element on the wrapper. Fixes #574


marksreeves:
Commit #fb4263f fixes #576

Issue #646, opened 2016-05-27
ActionConstraint not honored


ricksbrown:
This is due to the way [wc/ui/table/common](https://github.com/BorderTech/wcomponents/blob/charlie/wcomponents-theme/src/main/js/wc/ui/table/common.js) is passed around between table modules. Its reusable widgets are modified in one module which affects every other module.

For example in `wc/ui/table/pagination`: 

``` 
PAGINATION_BUTTON = common.BUTTON
...
PAGINATION_BUTTON.descendFrom(PAGINATION_CONTAINER);
 ```

Or this in `wc/ui/table/sort`:

```
THEAD = common.THEAD
...
THEAD.descendFrom(SORTABLE_TABLE, true);
```

Or even this looks sus:

```
SORT_CONTROL = common.TH.extend("", {"aria-sort": null})
...
SORT_CONTROL.descendFrom(THEAD);
```

Edit: I think the SORT_CONTROL one is probably OK.

ricksbrown:
Closed via #647

PR #647, opened 2016-05-27
Table constraints enforced

ricksbrown:
Well... Due to operational constraints this was a little more hurried that I would have liked however it should do the trick and:

* Fix existing modifications to common table widgets
* Prevent new modifications from happening via `descendFrom`

Probably the whole common table widget thing needs to be redone.

Closes issue #646


ricksbrown:
@marksreeves This is directly connected to some of your recent work. Would you mind confirming that this fix looks right - i.e. the modification of the common table widgets was, indeed, an unintentional side effect.
If that's all good feel free to merge this PR.

marksreeves:
Looks alright to me.

PR #699, opened 2016-06-22
WTextArea sanitization

### LINKED ISSUE #620: WTextArea sanitization
We added the ability to convert WTextArea to a rich text input. There should be a default implementation of:
* a XML output sanitizer to allow HTML entered into a WTextArea to be sent back to that WTextArea without breaking XML - this is particularly pertinent to HTML entities which are not XML entities (which is most of them); a
* a HTML input sanitizer which disallows some elements (at the very least SCRIPT, OBJECT, and preferably to disallow common extension elements not part of the spec eg FONT, APPLET).

@jonathanaustin comments?

marksreeves:
* Added a mechanism to make HTML entities safe in XML.
* Added a wrapper for AntiSamy and a default config suitable for tinyMCE.
* Updated WTextArea.setData to use the HTML entity unescaper and AntiSamy for input sanitization if in rich text mode.
* Updated WText.setData to use the HTML entity unescaper.
* Updated the XSLT build to retain inter-node whitespace in the html namespace.
* Updated the XSLT for WTextArea in the read-only state to allow for nested HTML.
* Added a null transform for ui:rtf which was missing.
* Updated TextAreaExample and WTextExample to show impact of the new escape/sanitize mechanisms.
* Updated wcomponents-core dependencies for apache commons to upgrade commons-lang 2.3 to commons-lang3 3.4 and add dependency for AntiSamy.

Fixes #620


marksreeves:
@jonathanaustin could you please take a look at the Java bits of this before I merge it.

wcomponents-core/pom.xml
```diff
@@ -83,10 +83,12 @@
...
 		</dependency>
 
+
 		<dependency>
-			<groupId>commons-lang</groupId>
-			<artifactId>commons-lang</artifactId>
-			<version>2.3</version>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
+			<version>3.4</version>
```
jonathanaustin: The biggest change from version 2.3 to 3.4 is the package name is ".lang3" so it can exist with older versions.

I noticed the "handler" bars library we are using is already bringing in version 3.1, and the "commons config" library is still bringing in the old 2.3 version.

If we are upgrading commons-lang. It might pay to look at all our libraries, or at least the apache commons.

wcomponents-core/pom.xml
```diff
@@ -203,6 +205,12 @@
 			<version>2.6.2</version>
 		</dependency>
 
+		<!-- Required for HTML input sanitization of WTextArea -->
```
jonathanaustin: We should look at doing some performance testing around this library.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WText.java
```diff
@@ -78,6 +80,25 @@ public void setText(final String text, final Serializable... args) {
 	}
```
jonathanaustin: I am concerned the direction we are sanitizing (ie on the get or set). Many times WText is used with a "bean property" and "setData" is never called.

Maybe we should do the sanitizing on getData?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTextArea.java
```diff
@@ -65,6 +69,30 @@ protected TextAreaModel getComponentModel() {
 	}
 
 	/**
```
jonathanaustin: Same concern as for WText for the direction we are sanitizing.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTextAreaRenderer.java
```diff
@@ -60,7 +58,10 @@ public void doRender(final WComponent component, final WebXmlRenderContext rende
 			xml.append("<ui:rtf />");
 		}
 
-		xml.appendEscaped(textArea.getText());
+		String textString = textArea.getText();
```
jonathanaustin: Just to make sure I understand....

The HTML sanitizing will make it safe to include in XML?

Why we are not escaping the text for "ReadOnly" and "Rich Text" field? How is it different to "not readonly"?

Is it dangerous to not escape and allow "malicious" script through?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/HtmlSanitizer.java
```diff
@@ -0,0 +1,100 @@
+package com.github.bordertech.wcomponents.util;
```
jonathanaustin: Might not seem necessary, but a bit of a naming convention we have is that any "util" classes are suffixed with "Util".

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/HtmlSanitizer.java
```diff
@@ -0,0 +1,100 @@
...
+
+	/**
+	 * Apply sanitization rules to a HTML string.
+	 *
+	 * @param input the (potentially) tainted HTML to sanitize
+	 * @return sanitized HTML
+	 * @throws NullPointerException thrown if AntiSamy instance is not correctly instantiated.
+	 * @throws Exception thrown if the AntiSamy scan fails.
+	 */
+	public static String sanitize(final String input) throws NullPointerException, Exception {
```
jonathanaustin: Might be better to throw the actual exceptions that are being caught on the Method....ScanException and PolicyException.

You could also create your own Exception "AnitSamyNotConfiguredException" to handle the null pointer exception. Or just throw a "System Exception" which does not force the applications to handle. Depends how you want to handle it

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/StringEscapeHTMLToXML.java
```diff
@@ -0,0 +1,39 @@
+package com.github.bordertech.wcomponents.util;
```
jonathanaustin: Do you think this method could just be added into the HtmlSanitiser util class? It is not really "extending" the apache class.

jonathanaustin:
@marksreeves have added some comments.....

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTextAreaRenderer.java
marksreeves: This has nothing to do with malicious HTML or script. It unescapes HTML character entities such as &bul; back to their character. These entities are not permitted in XML whereas the Characters are because we serve it'd-8. This was the proper fix we should have done years ago instead of putting in a doctype and telling people not to use these entities.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/HtmlSanitizer.java
marksreeves: OK, will rename.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/StringEscapeHTMLToXML.java
marksreeves: I am not sure what you mean here. By extending StringEscapeUtils we can use this class to also call the static final methods in StringEscapeUtils (such as escapeXML10). It is also a different thing from a sanitizer: this does not sanitize, it unescapes: for example `&bul;` to `•`

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WText.java
marksreeves: Roger that.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTextArea.java
marksreeves: No problem. I built this to sanitize input from a WTextArea in rich text mode and this is the lowest impact time to do that level of sanitization (it also means only cleaned HTML goes into storage).

I suppose I shouldn't trust HTML coming from a data source any more than I should trust it coming from a "human".

Maybe we have to sanitize both ways for WTextArea: incoming if richText & editable and outgoing for all richText?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTextAreaRenderer.java
marksreeves: To further clarify: anything which is not WTextArea with rich-text **and** read-only or is not WText **and** unencoded gets XML encoded. 

In all cases where XML encoding (actually escaping: another problem with our API) happens `&bul;` would become `&amp;bul;`. `&amp;` is XML safe, `&bul;` is not. Therefore we only need to worry about HTML character entities in the two cases mentioned above. We unescape HTML entities which are not XML entities because if we did not then these entities would cause XML failure (or we could be really perverse and add a doctype for **all** entitities 😉). 

I would prefer to get rid of XML completely but that is not an option in the time frames in which we **have** to deliver a means to make WTextArea in rich-text mode safe to implement.

Issue #689, opened 2016-06-16
Simplify theme overrides and protect integrity of UI components


marksreeves:
There are too many options for theme building which can leave the structure of the components in a mess or reduce the accessibility aims of WComponents. The way the themes are built allows any part of the theme to be overridden. We should keep this but make simple overrides easier and discourage (possibly merely through comment convention) theme-level overrides of structural stuff.

Any suggestions for achieving this are welcome.

marksreeves:
The XSLT changes in PR #1084 moved this a long way towards completion. It will not be complete whilst we allow XSLT over-ride in themes.

PR #733, opened 2016-07-06
Sass fixes

marksreeves:
Sass optimisations for #689:

* Changed the mechanism to allow/disallow generic button font metric
inheritance from overriding a Sass file to changing a var.
* Fixed some redundant CSS around buttons and disabled elements
* Added a mechanism to optionally apply “invite” effects to buttons
using a Sass var
* Allow generic error state background colours to be switched off.

Sass bugs:

* Fixed a bug which resulted in inconsistent styling of labels for
checkboxes and radios under some conditions.


marksreeves:
Looks like a npm problem on Travis at the moment. Still investigating why the build is failing with a missing node module. Cannot reproduce locally with a clean node install.

PR #761, opened 2016-07-15
AntiSamy and cater for override

jonathanaustin:
Extension to #620.

Some refactor to allow projects to override AntiSamy functionality. Components using AntiSamy now have protected methods `sanitizeInputText(text)` and `sanitizeOutputText(text)`.

PTAL @Joshua-Barclay @marksreeves


Joshua-Barclay:
Good change - I think this significantly improves usability and mutabiity as a framework utility.

Issue #658, opened 2016-06-03
WComponents Configuration


Joshua-Barclay:
The Config and DefaultInternalConfiguration classes in WComponents could be enhanced to be more usable for applications.

1. There is no way of knowing which parameters have been configured for WComponents. At a minimum the configuration should be logged once it's established.
2. Applications cannot specify their own files and/or name files to match a generic pattern (e.g. `wcomponents-*.properties`)
3. Applications are unable to configure their own provider - ideally there would be a way to have their own configuration provided specified (like SLF4J)

Item 1 above is a significant priority. Item 2 and 3 are less important but do provide substantial value.

marksreeves:
Is this in any way related to #249? Would, for example, the provision of an over-rideable provider - as per item 3 above - meet the base need of #249? My reading of #249 is that it is not about an annotation override of configuration per-se but a more _common_ (I am loathe to use the word 'standard') method to provide config with @JohnMcGuinness using annotation as an example of such a means.

Joshua-Barclay:
Issues 2 and 3 resolved by ServiceLoader in #863 

Remaining issue is that the WComponents parameters are not logged.

Issue #249, opened 2015-12-07
WComponents configuration should move towards more 'standard' methods of configuration


JohnMcGuinness:
I think most JEE developers are familiar with the concept of configuring their application via XML or Annotations. Where the annotation config can be overriden by XML. 

I think, but I am happy to be disproved, that few if any web frameworks use properties files as their primary means of configuration.

I would like to propose moving WComponents in the direction of favouring Annotations/XML as its primary means of configuration.

As a first step I would like to add the WServletConfiguration annotation. An example is below:

```
@WServletConfiguration( ui = MyApp.class)
public class MyWSerlet extends WServlet {
}
```

One benefit of this is that it is type safe. At compile time if the developer enters a value for `ui` that is not a subclass of `WCompnent` or maybe `WApplication` their IDE will tell them.

Also it obviates the need for the developer to be aware of `UIRegistry` and implement the `getUI()` method.

But above all it's simple, and makes it easier for a beginner to get started. I like API's that are simple.

WDYT @jonathanaustin @ricksbrown @marksreeves @Joshua-Barclay

jonathanaustin:
Totally agree that we should look at moving in this direction.

An item we have had on our "todo" list for a long time is making WServlet easier to use. One way we thought was to make WServlet take an init parameter that defines the "application UI" class. The getUI method could look for this parameter, create an instance of the class and "register" it.

Using this parameter also means a project would not even have to create their own WServlet class.

But it would be awesome if we could create an annotation that projects can use on their WApplication class that setups the web servlet and registers it at the same time. Not sure if this could be done as the @WebServlet annotation does expect the class to extend HttpServlet. Maybe the new annotation could somehow link out to the WServlet class and set the init parameter to the WApplication class.

But...Annotations are definitely the way to go though.

PR #784, opened 2016-07-19
Refactor configuration into ConfigurationProperties

Joshua-Barclay:
General refactoring work to improve consistency and maintainability of WComponents Configuration.

Generally related to #658 and #249, but does not directly resolve either.

@JohnMcGuinness fyi


JohnMcGuinness:
This is a really good way to centralize and simplify WComponents configuration.

I was wondering about setting configuration values, usually these will be read from properties files, but would it be worth adding a set(property) method to enable setting properties at runtime, or should configuration be considered as immutable.

WDYT @Joshua-Barclay

Joshua-Barclay:
ConfigurationProperties is intended for internal WComponents configuration only. WComponents does not set properties using code, so it's not relevant to the ConfigurationProperties class. WComponents does set properties in test classes, but using `Configuration.setProperty()` with the property declaration from `ConfigurationProperties` should be adequately maintainable. 

In more general terms, although the Configuration class has a `setProperty` API, Configuration should generally be immutable. I can't think of a valid scenario for setting a property in Java other than unit tests.

PR #796, opened 2016-07-25
Subordinate rule proccessing

### LINKED ISSUE #780: Subordinate rule processing - read-only controls
Components which implement SubordinateTrigger when in a read-only state (`component.isReadOnly() == true`) should process their subordinate rules _without_ reference to their disabled state.

A control in a readOnly state has a value which was immutable in the current view. It is a nonsense to `setDisabled(true)` on a control which is in a read-only state but nothing in the API prevents this so we have to make sure we ignore the disabled state when doing server-side subordinate rule processing on any control in a read only state.

This is a necessary precursor to #261.

marksreeves:
Updated comparison execution such that disabled is only assumed false
if the control is not also disabled. Fixes #780.

@jonathanaustin PTAL

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/subordinate/AbstractCompare.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/subordinate/AbstractCompare.java
@@ -88,7 +88,7 @@ public abstract class AbstractCompare extends AbstractCondition {
 	@Override
 	protected boolean execute() {
 		// Disabled triggers are always false
-		if ((trigger instanceof Disableable) && ((Disableable) trigger).isDisabled()) {
+		if ((trigger instanceof Disableable) && ((Disableable) trigger).isDisabled()  && !(trigger instanceof Input && ((Input) trigger).isReadOnly())) {
 			return false;
 		}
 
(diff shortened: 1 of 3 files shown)


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/subordinate/AbstractCompare.java
```diff
@@ -107,7 +107,7 @@ protected boolean execute() {
 	@Override
 	protected boolean execute(final Request request) {
 		// Disabled triggers are always false
-		if ((trigger instanceof Disableable) && ((Disableable) trigger).isDisabled()) {
+		if ((trigger instanceof Disableable) && ((Disableable) trigger).isDisabled() && !(trigger instanceof Input && ((Input) trigger).isReadOnly())) {
```
jonathanaustin: This is good. `Readonly` takes precedence over `Disabled`. This applies to the `Trigger`, but we should also make sure the server side logic is the same as the client for `actions`.

Should `mandatory/optional`, `hide/show`, `enable/disable` actions not be applied to `targets` that are `readonly` or `disabled` (except for the enable/disable of course). What does the client do for these actions?

marksreeves:
The knowledge of what is hide-able, disable-able and mandate-able is encapsulated in a utility so we can call `setDisabled` on anything knowing that if it is not disableable the utility will take care of it.

What this means practically is that when a component is in its readOnly state the only subordinate actions which apply are show/hide. Enable, disable, mandatory and optional are all ignored.

PR #806, opened 2016-07-26
Fix dialog position, responsive UI updates.

### LINKED ISSUE #805: DialogFrame initial position may be incorrectly calculated
When a dialog is opened its position is calculated unless top and left are explicitly set. At the moment the test is for either top _or_ left whereas both are required to correctly position the dialog frame and if one is not set then the dialog will need to be positioned.

To view the issue in the wcomponent examples LDE:

1. open the WDialog example;
2. resize the browser so that the viewport width is greater than the width of the "enormous dialog" but the height is less than the dialog's height;
3.  Open the "enormous dialog";
4. close the dialog;
5. Increase the viewport height to be greater than the height of the "enormous dialog";
6. reopen the "enormouse dialog" (do not reload the view: it is an immediate opening dialog.

See that the dialog does not open in the correct position.
line 393:
``` javascript
if (!(obj.top || obj.left || obj.top === 0 || obj.left === 0)) {
```

should be:
``` javascript
if (!((obj.top || obj.top === 0) && (obj.left || obj.left === 0))) {
```

marksreeves:
* Updated dialogFrame.js to fix positioning bug. Found another issue in `wc/ui/dialog` which could result in an incorrect position being preserved between opening (negative left or top). Updated example to test fix. Fixes #805.
* Resize event driven UI manipulation saw a scattering of toggle points so I created a utility module to manage them all. Should be easier to keep in sync with the Sass this way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Fixed a bug in `com.github.bordertech.wcomponents.subordinate.AbstractCompare` which resulted in Subordinate controls returning an incorrect value if the control was in a read-only state #780.
 * Fixed a newly introduced bug which caused textareas to fail to accept newlines in IE11 #785.
 * Fixed several IE CSS issues.
+* Fixed bug which could result in dialogs being mis-positioned #805.
 
 ## Enhancements
 * Added mechanism to convert tabsets to accordions on small screens #783.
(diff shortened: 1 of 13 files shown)


marksreeves:
Fixes #808, #809

PR #818, opened 2016-08-01
WDialog launch 407

### LINKED ISSUE #407: Allow WDialog to be triggered by an AjaxTrigger
Is it possible to make the trigger for a WDialog an AjaxTrigger rather than a WButton? Then we could do client-side triggering of dialogs using WMenuItem.

This is the WDialog constructor for wcomponent-core-1.0.3 :
          public WDialog(final WComponent content, final WButton trigger) {
                   add(holder);

                   if (content != null) {
                             setContent(content);
                   }

                   this.trigger = trigger;

                   if (trigger != null) {
                             add(trigger);
                   }
          }

marksreeves:
Updated WDialog to allow any component which implements AjaxTrigger to be a launch component. The launch component is not added to the WDialog in the way WButton was. Deprecated the constructor which accepts a WButton arg. Fixes #407


jonathanaustin:
LGTM

PR #863, opened 2016-08-19
Provide reliable configuration override mechanism

Joshua-Barclay:
The current mechanism requires consumers to hook into `Config.setConfiguration()` at the earliest point they can. This is incredibly fragile as it relies on ClassLoading order, and in fact there may be some properties retrieved prior to any hook being available.

The new mechanism uses the J2SE ServiceLoader pattern to search for `ConfigurationLoader` implementations within the classpath. These implementations then provide their specific `Configuration` instance which is combined with the WComponents `DefaultInternalConfiguration` using a `CompositeConfiguration`.


Joshua-Barclay:
@jonathanaustin 
Please review this PR and merge if you are happy with it. Happy to discuss further.

jonathanaustin:
LGTM

PR #864, opened 2016-08-22
Responsive CSS, WButtonRenderer

marksreeves:
* Added a mechanism to apply different sized padding to automatically padded components when viewed on narrow viewports. This brings these padded components. such a WSections and WPanel TYPE BOX, into line with margin and h/vgap.
* Updated WButtonRenderer so it does not throw a SystemError if the WButton has a toolTip or accessibleText.
* minor change to the WTextArea example to make rich text a bit nicer to use.

@jonathanaustin PTAL at the WButtonRenderer change


jonathanaustin:
LGTM

PR #865, opened 2016-08-22
Multiple HTML Classes 856

### LINKED ISSUE #856: Improve mechanism to add multiple HTML class attribute values
AbstractWComponent has a member `private String htmlClass` which holds a single String which can be used to add additional class attribute value(s) to a component's output HTML. To add new value(s) we currently have to get the current value (if any), append a space then append the additional value(s); then use this concatenated String as the argument for `setHtmlClass(String)`. If one or more of the new values are from the `HtmlClassProperties` enum then we have to manually call `toString()` on those before we can use them. 

This should be made easier by either:

1. providing methods to do this without having to get and append spaces; or
2. by using a List<String> instead of a simple String member.

The first is quick (and a bit dirty) the second has a performance hit but can then also be leveraged to prevent the same value being added more than once and to remove value(s).

@jonathanaustin WDYT?

marksreeves:
* COnverted Serializable HtmlClassName to a HashSet<String>.
* Added methods addHtmlClass(String), addHtmlClass(HtmlClassProperties), removeHtmlClass(String) and
removeHtmlClass(HtmlClassProperties)

Fixes #856.

@jonathanaustin PTAL

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
@@ -19,9 +19,9 @@ import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.commons.logging.Log;
@@ -1695,9 +1695,9 @@ public abstract class AbstractWComponent implements WComponent {
 	 * {@inheritDoc}
 	 */
 	@Override
-	public HashSet getHtmlClassList() {
+	public Set getHtmlClasses() {
 		ComponentModel model = getComponentModel();
-		return model.getHtmlClassList();
+		return model.getHtmlClasses();
 	}
 
 	/**
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
@@ -19,6 +19,7 @@ import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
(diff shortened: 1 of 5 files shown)


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
```diff
@@ -1707,6 +1691,31 @@ public String getHtmlClass() {
 		return I18nUtilities.format(null, model.getHtmlClass());
 	}
 
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public HashSet getHtmlClassList() {
```
jonathanaustin: Where possible, it is good to use the interface "Set" instead of an implementation like "HashSet".

The name of the method is a little bit misleading....What if we call it getHtmlClasses().

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/ComponentModel.java
```diff
@@ -182,7 +182,7 @@
 	/**
 	 * Adds extra value to the HTML class of the output component.
 	 */
-	private Serializable htmlClass;
+	private HashSet<String> htmlClassList;
```
jonathanaustin: This can also be Set<String>

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/ComponentModel.java
```diff
@@ -537,25 +537,95 @@ protected void setAccessibleText(final String text, final Serializable... args)
...
+	protected HashSet getHtmlClassList() {
+		return this.htmlClassList;
+	}
+
+	/**
+	 * Remove a value from the HTML class name list.
+	 * @param className the value to remove
+	 */
+	protected void removeHtmlClass(final String className) {
+		if (!(this.htmlClassList == null || Util.empty(className))) {
```
jonathanaustin: I would tend to code this without the wrapping negative....`this.htmlClassList != null && !Util.empty(className)`...but that is just me.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/ComponentModel.java
```diff
@@ -537,25 +537,95 @@ protected void setAccessibleText(final String text, final Serializable... args)
...
+
+	/**
+	 * @param text The HTML class name text to set.
+	 */
+	protected void addHtmlClass(final String text) {
+		if (!Util.empty(text)) {
+			if (this.htmlClassList == null) {
+				this.htmlClassList = new HashSet<>();
+			}
+			this.htmlClassList.add(text);
```
jonathanaustin: Do you think it is worth 'trimming` the value?

Issue #803, opened 2016-07-26
Provide access to HTTPRequest


marksreeves:
The Request interface does not provide access to the backing request. We have had to modify this several times to provide access to request properties (such as specific headers) and we have a need to provide information about the user agent, platform etc as exposed in the request. 

We could expose all headers individually but this would not account for custom headers and seems a bit of a waste when we could expose the backing request.

marksreeves:
We have also promised some consumers that we could provide a helper to determine if a request is coming from a mobile platform. This could be ascertained (as far as it is possible to do so) from the backing request but we should add a helper that does the heavy lifting so that multiple consumers do not have to do the same check for themselves.

jonathanaustin:
The `Request` interface is a throw back to `Portal`. WComponents had to process two different types of requests, a `PortalRequest` or `HttpServletRequest`. As we no longer support `Portal` we could look at removing this abstraction layer.

This would be a major API change so would have to wait until Q4.

In the mean time, we can add a method to `ServletRequest` to provide access to the backing `HttpServletRequest`.

We can also provide a new helper method in `ServletUtil` to determine the user's device type.

marksreeves:
Sounds like a plan. I have rescheduled.

jonathanaustin:
There is value in maintaining the abstraction layer of our own `Request` and `Response` layers as `WComponents` may not always be running in a `Servlet` environment. It could be running in a non Java JVM such as `JRuby`, `Jython`, `Closure` or `Scala`.

Therefore, the lifecycle processing of `WComponents` should not be dependant on `HttpServletRequest` and `HttpServletResponse`.

marksreeves:
I am closing this as resolved. @jonathanaustin did as much as we can, and solved the original enhancement request, by creating `ServletRequest#getBackingRequest()`.

@jonathanaustin feel free to re-open if you feel this is insufficient.

PR #898, opened 2016-08-31
Device Type

jonathanaustin:
Addresses #803.

Provide access to the backing `HttpServletRequest` in `ServletRequest`.

A new helper method in `ServletUtil` to determine the user’s device.

### DIFF
--- a/wcomponents-core/src/test/java/com/github/bordertech/wcomponents/servlet/ServletUtilDeviceType_Test.java
+++ b/wcomponents-core/src/test/java/com/github/bordertech/wcomponents/servlet/ServletUtilDeviceType_Test.java
@@ -63,9 +63,15 @@ public class ServletUtilDeviceType_Test {
 
 	@Test
 	public void testDeviceType() {
+
+		// HttpServletRequest
 		MockHttpServletRequest request = new MockHttpServletRequest();
 		request.setHeader("User-Agent", userAgent);
 		Assert.assertEquals("Incorrect device type for " + desc, expectedDeviceType, ServletUtil.getDevice(request));
+
+		// ServletRequest
+		ServletRequest req = new ServletRequest(request);
+		Assert.assertEquals("Incorrect device type for " + desc + " using ServletRequest", expectedDeviceType, ServletUtil.getDevice(req));
 	}
 
 }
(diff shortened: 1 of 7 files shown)


jonathanaustin:
PTAL @marksreeves @ricksbrown

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/servlet/ServletUtilDeviceType_Test.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.runners.Parameterized;
+
+/**
+ * Parameterized test for device types in {@link ServletUtil}.
+ *
+ * @author Jonathan Austin
+ * @since 1.2.5
+ */
+@RunWith(Parameterized.class)
+public class ServletUtilDeviceType_Test {
```
ricksbrown: I like the tests. 

Do you think we should add UC user agent string to the list?

Also some of those user agent strings look like templates, for example `<Build Tag etc.>` do you think we should actually file them out or turn them into sprintf strings?

ricksbrown:
LGTM

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/servlet/ServletUtilDeviceType_Test.java
```diff
@@ -0,0 +1,77 @@
...
+
+	private static final String MOBILE_SAFARI = "Mozilla/5.0 (iPad; CPU OS 9_3_3 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13G34 Safari/601.1";
+
+	private static final String CHROME_WIN_10 = "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36";
+
+	private static final String CHROME_MOBILE = "Mozilla/5.0 (Linux; <Android Version>; <Build Tag etc.>) AppleWebKit/<WebKit Rev> (KHTML, like Gecko) Chrome/<Chrome Rev> Mobile Safari/<WebKit Rev>";
```
marksreeves: The parts of the user agent strings in angle bracket are, indeed, templates. They are not _really_ important for the purposes of these tests. If you want to replace them with some actual common user agent strings see http://www.useragentstring.com/pages/useragentstring.php

marksreeves:
LGTM

jonathanaustin:
We can definitely add more User Agent Strings as we go. I will merge the change.

PR #905, opened 2016-09-05
XSL Interceptor special case

ricksbrown:
Facilitates zero configuration for any useragent that wants to opt itself out of XSLT.

@jonathanaustin PTAL


jonathanaustin:
LGTM

PR #917, opened 2016-09-13
Menu negative margin

### LINKED ISSUE #916: "Docking" menu bars
When a WMenu of `Type.BAR` is the first child of a WPanel of `Type.CHROME` or of `Type.ACTION` (ie a WPanel with exposed title) _or_ is the first child of a WPanel child of WSection; **and** that WPanel does not have a Layout applied then a negative margin is applied to make the menu "dock" to the header bar.

It is not always feasible to make the menu the first child a panel with no layout and have that be the relevant first child of a section.

So I propose to add a class value which will apply the same neg margin to any WMenu of Type.BAR and will take into account the theme set values of the section padding and any responsive design reduction in this (as determined by the Sass in wc.common.section.scss).

marksreeves:
Added class `wc-neg-margin` which, if applied to a WMenu of `Type.BAR`
will give it a negative margin on the top, left and right which will
cause it to dock to a WSection header or the title bar of WPanel types
`CHROME` and `ACTION`.

Resolves #916


marksreeves:
@jonathanaustin this will solve your issue of having to deeply nest a menu bar.

Issue #990, opened 2016-11-21
WMultiFileWidget sends XML to client when server XML is enabled


marksreeves:
When uploading a file using WMultiFileWidget the file XML is sent to the client even when the server XSLT flag is set.

jonathanaustin:
This was happening because the file upload functionality occurs via the `targeted` life cycle and the response does not go through the XML transformation interceptor.

It has been fixed so the `upload` functionality now goes via the `AJAX` lifecycle which means the `response` will be transformed.

In looking at this issue I also found the `WrongStepAjaxInterceptor` had a similar problem where it was generating an XML response that was not being transformed as it was above the `TransformXMLInterceptor` in the chain. This has also been fixed by putting the `TransformXMLInterceptor` at the top of the chain for both `AJAX` and `Page submit` chains.

The `WrongStepAjxInterceptor` also had to be modified to render its response in the `render` phase instead of the `action` phase to make sure it was transformed.

PR #991, opened 2016-11-21
XSLT 2 - remove JavaScript XSLT processor

marksreeves:
JavaScript XSLT processor cannot process XSLT 2.


marksreeves:
Required as part of fix for #990

PR #992, opened 2016-11-21
Ajax trigger registry

marksreeves:
All of the following form part of removing ajax-alias when it is not
necessary. This is part of templating improvements prior to removing
XSLT.

- Removed unused ajax alias.
- Updated mechanisms to avoid ajax when selecting an option in a SELECT element when that select is not itself an ajax trigger (descendant of WShuffler or WMultiSelectPair)
- Updated all components which do manual ajax calls so that they no longer rely on an alias being present.
- Ajax alias no only placed on components which can determine it directly from the component’s template and not from a cross reference to a list of ajax triggers.


marksreeves:
Note: this also addresses some of the issues which cause #990 to fail.

Issue #639, opened 2016-05-19
Farewell XSLT?


ricksbrown:
This issue discussed the move away from XSLT, specifically client side XSLT but by extension, in the interests of simplicity (removing unnecessary layers), removing the XSLT completely.

The use of client side XSLT is dependent on ongoing support by major browsers, however browser vendors have lost interest in XML / XSLT. 

It is **common** for us to come across issues that only occur on HTML pages generated by XSLT, for example:

- Mobile Safari (iPhone) recently stopped being able to process our XSLT #126 (this happened right when WComponents was being evaluated for use in a system that needed to support iPhone).
- Terrible memory leaks in IE8 [when page generated by XSLT](https://support.microsoft.com/en-us/kb/2032595)
- Firefox performance API completely broken in pages generated by XSLT [but XSLT gets low priority](https://bugzilla.mozilla.org/show_bug.cgi?id=804069#c8):

> Honestly, I don't know of anyone not working on something more important. XSLT is not seeing a whole lot of usage on the web.

- Original Android browser did not do XSLT at all.
- Chrome's implementation of Web Components [throws errors if the page is generated by XSLT](https://code.google.com/p/chromium/issues/detail?id=551089).
- Windows Phone 8.1 lost all ability to do client side XSLT #141 
- UC Browser on Android does not do XSLT at all.
- There is even discussion on the [Chromium developer forum](https://groups.google.com/a/chromium.org/d/msg/blink-dev/zIg2KC7PyH0/Ho1tm5mo7qAJ) which raises the possibility of altogether removing support for client side XSLT from Blink.
- The webkit implementation of XSLT is buggy and referred to as a "hack" by the original implementer:

> I am the original author of XSLTProcessor for WebKit, as well as much of WebKit's XSLT support (read: hacks). 

Is XSLT a powerful transformation engine that ships with every major browser out of the box? Yes.
Has it seen massive uptake and support? No.

It's time to demote it, move it to the server and start replacing it with plain old HTML and perhaps [HTML Custom Elements](https://www.w3.org/TR/custom-elements/) (which solves all the problems that originally drove us to XSLT).

See also #174

marksreeves:
[...]

PR #1035, opened 2016-12-19
Force xslt

marksreeves:
XSLT must take place in WComponents, not in a browser. This requires that we remove the ability to avoid the XSLT interceptor. This change removes the use of the PARAM without removing any of the public API, thereby allowing us to deprecate the public API for removal in v2.

Forms part of #639.

@jonathanaustin PTAL


wcomponents-core/src/test/java/com/github/bordertech/wcomponents/container/TransformXMLInterceptor_Test.java
```diff
@@ -59,26 +59,12 @@ public static void tearDownClass() {
 	}
 
 	/**
```
jonathanaustin: You could keep this test, with setting the parameter to false and prove it still transforms.

wcomponents-lde/pom.xml
```diff
@@ -43,8 +43,15 @@
 			<artifactId>junit</artifactId>
 		</dependency>
```
jonathanaustin: wcomponents-core has a "dummy" xslt for testing the "transform". What do you think of using something similar for this test and then it wont be dependant on the "theme"...

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/container/TransformXMLInterceptor_Test.java
marksreeves: Not really happy with this. It is OK to prove the change has been made but not really useful after the initial pull.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/ConfigurationProperties.java
```diff
@@ -323,6 +323,7 @@
 
 	/**
 	 * Whether to perform server-side XSLT.
```
marksreeves: Yep, ok.

wcomponents-lde/pom.xml
marksreeves: This change was in line with wcomponents-examples. I am not really keen on having a dummy XSLT. We have a process which relies on an XSLT processor and we build XSLT so why not use the one we already have?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/TransformXMLInterceptor.java
```diff
@@ -93,7 +93,7 @@
 	 */
 	@Override
 	public void preparePaint(final Request request) {
-		doTransform = isPerformTransform();
+		doTransform = true;
```
marksreeves: Yeah, OK. It is unlikely that the header would change between requests within the same session.

marksreeves:
Schema changes to fix #1036.

PR #1040, opened 2016-12-23
Fix excessive TransformXMLInterceptor WARN logging

### LINKED ISSUE #1039: Excessive WARN logging for XSLT Illegal HTML Characters
TransformXMLInterceptor has a temporary workaround configuration option to strip illegal HTML characters prior to Saxon transforming the XML.

In order to demonstrate the severity of the issue hidden by this workaround, a warning message was added for each transformation:
```java
public void paint(final RenderContext renderContext) {
...
		if (isAllowCorruptCharacters() && !Util.empty(xml)) {

			// Remove invalid HTML characters from the content before transforming it.
			LOG.warn("Allowing corrupt characters.");
			xml = removeCorruptCharacters(xml);
		}
```

This fires a WARN message on every single request, whether they contain illegal characters or not.
This is not an appropriate level of logging.

The WARN message should instead be logged when an illegal character is stripped.

Joshua-Barclay:
Closes #1039 

@jonathanaustin PTAL


jonathanaustin:
LGTM

PR #1080, opened 2017-02-03
XSLT clean up

marksreeves:
I have undertaken a large rationalisation of the XSLT as another step towards #639. This has involved some changes to Java renderers so that cross-element lookups in the XSLT could be removed. Any property which could not be calculated if the element was in an AJAX response has likewise been moved to the Java renderer or into JavaScript if it was purely client-side.

There is still some optimisation to do, especially around `WLabel` rendering and position.

The precis of Java changes is as follows:

## API Changes

**NOTE** all Java API changes are backwards compatible.

- `WMenuItem` & `WSubMenu`: added public method `boolean isTopLevelItem()` required by the renderer of each
- `WSubMenu`: `getSelectability()`, `setSelectability(Boolean)`, `setSelectable(boolean)` and `setSelected(boolean)` are all deprecated (always ignored in client UI).
- `WMenuItemGroup`: `isDisabled()` will return true if the group is nested in a `WSubMenu` or `WMenu` which is disabled . This brings it into line with WMenuItem and WSubMenu and provides for no UI change as previously this calculation was in the client UI.

### Margins, gaps and spaces

Margins and intra-component spaces (`hgap`, `vgap`) were historically set as `int` but were output into the UI as an enumerated list of options based on nearest-high-neighbour calculations. This was done to improve consistency and enhance responsive design. These calculations have been removed from the client UI to Java and all spaces (`Margin` and intra-component spaces) have been changed from int to values of an enum. A temporary backwards compatibility layer has been provided and all former `int` based constructors and getters have been deprecated. Note that the spaces are immutable and so it is not expected that the getters would be used externally - they were public only for use in the components' renderer.

- Added new utility class `SpaceUtil`.

  - Used to manage consistent inter- and intra- component spaces by converting `Margin`, and `[h|v]?gap` from int to enum values.

  - enum values are

    - `ZERO` used for removing any theme-inbuilt spacing (such as in Headings);
    - `SMALL` for small spaces;
    - `MEDIUM` for medium spaces;
    - `LARGE` for large spaces;
    - `XL` for very large spaces;

    in addition using `null` when setting a gap or `Margin` will remove any previously set space and allow the theme defaults to apply. In this way it is possible to set a margin (for example) on only one side of a component without explicitly setting `ZERO` on the other sides. This provides 1:1 compatibility with the previous theme-determined spacing whilst still allowing for an **explicit** `ZERO` space.

  - The utility has temporary helpers to convert `int` gaps/margins to the enum.

- `Margin`
  - All current `int`-based constructors are deprecated and use helpers in `SpaceUtil` to convert int spacing to the enum. The values used for the conversion points are as per the former conversion client UI so there will be **no visible change** in gaps or margins.
  - `int getAll()` deprecated in favour of `SpaceUtil.Size getMargin()`
  - `int getNorth()` deprecated in favour of `SpaceUtil.Size getTop()`
  - `int getEast()` deprecated in favour of `SpaceUtil.Size getRight()`
  - `int getSouth()` deprecated in favour of `SpaceUtil.Size getBottom()`
  - `int getWest()` deprecated in favour of `SpaceUtil.Size getLeft()`  

- `BorderLayout`, `ColumnLayout`, `GridLayout`:
  - All current `int`-based constructors are deprecated and use helpers in `SpaceUtil` to convert int spacing to the enum. The values used for the conversion points are as per the former conversion client UI so there will be **no visible change** in gaps or margins.
  - `int getHgap()` deprecated in favour of `SpaceUtil.Size getHorizontalGap()`
  - `int getVgap()` deprecated in favour of `SpaceUtil.Size getVerticalGap()`

- `FlowLayout`, `ListLayout`, `WList`, `WRow`:
  - All current `int`-based constructors are deprecated and use helpers in `SpaceUtil` to convert int spacing to the enum. The values used for the conversion points are as per the former conversion client UI so there will be **no visible change** in gaps or margins.
  - `int getGap()` deprecated in favour of `SpaceUtil.Size getSpace()`

## Enhancements

- `WAbbrTextRender` changed from XML to HTML with change in UI.
- `WDataTableRenderer` no longer renders any internal disabled states with no UI change as these were ignored in the UI.
- `WFieldErrorIndicatorRenderer` will not render with no change, previously ignored in UI.
- `WLabelRenderer` is aware of the nature and state of its labelled component and is able to correctly set the state of a `WLabel` in the UI without cross-reference to its component. This was required to fix several possible errors caused when a `WLabel` was in an AJAX response.
- `WMenuItemRenderer` and `WSubMenuRenderer` will only output the `accessKey` member if the component is at the top level of its `WMenu`
- `WTabGroupRenderer` will no longer paint `WTabGroup` as `ui:tabgroup` but pass through to its `WTabs` with no UI change as previously WTabGroups were ignored.
- Selenium test elements improved and extended.

@jonathanaustin @ricksbrown PTAL: all comments welcome.


marksreeves:
Have to sort out why Selenium has suddenly started failing.

PR #1084, opened 2017-02-05
Big XSLT refactor

marksreeves:
I have greatly simplified the XSLT as a stepping stone to #639 and to remove the cross-element calculations which were necessarily flawed in AJAX responses. 

This has necessitated some Java changes (all backwards compatible) as listed below; rather a lot of JavaScript changes to move some client-only calculations which used to be attempted in XSLT into JavaScript; and a few Sass/CSS changes.

The main change of note is in wcomponents-test-lib Selenium elements have been increased and improved.

# Java 

## API Changes (all backwards compatible)

- `WMenuItem` & `WSubMenu`: added public method `boolean isTopLevelItem()` required by the renderer of each
- `WSubMenu`: `getSelectability()`, `setSelectability(Boolean)`, `setSelectable(boolean)` and `setSelected(boolean)` are all deprecated (always ignored in client UI).
- `WMenuItemGroup`: `isDisabled()` will return true if the group is nested in a `WSubMenu` or `WMenu` which is disabled . This brings it into line with WMenuItem and WSubMenu and provides for no UI change as previously this calculation was in the client UI.

### Margins, gaps and spaces

Margins and intra-component spaces (`hgap`, `vgap`) were historically set as `int` but were output into the UI as an enumerated list of options based on nearest-high-neighbour calculations. This was done to improve consistency and enhance responsive design. These calculations have been removed from the client UI to Java and all spaces (`Margin` and intra-component spaces) have been changed from int to values of an enum. A temporary backwards compatibility layer has been provided and all former `int` based constructors and getters have been deprecated. Note that the spaces are immutable and so it is not expected that the getters would be used externally - they were public only for use in the components' renderer.

- Added new utility class `SpaceUtil`.

  - Used to manage consistent inter- and intra- component spaces by converting `Margin`, and `[h|v]?gap` from int to enum values.

  - enum values are

    - `ZERO` used for removing any theme-inbuilt spacing (such as in Headings);
    - `SMALL` for small spaces;
    - `MEDIUM` for medium spaces;
    - `LARGE` for large spaces;
    - `XL` for very large spaces;

    in addition using `null` when setting a gap or `Margin` will remove any previously set space and allow the theme defaults to apply. In this way it is possible to set a margin (for example) on only one side of a component without explicitly setting `ZERO` on the other sides. This provides 1:1 compatibility with the previous theme-determined spacing whilst still allowing for an **explicit** `ZERO` space.

  - The utility has temporary helpers to convert `int` gaps/margins to the enum.

- `Margin`
  - All current `int`-based constructors are deprecated and use helpers in `SpaceUtil` to convert int spacing to the enum. The values used for the conversion points are as per the former conversion client UI so there will be **no visible change** in margins.
  - `int getAll()` deprecated in favour of `SpaceUtil.Size getMargin()`
  - `int getNorth()` deprecated in favour of `SpaceUtil.Size getTop()`
  - `int getEast()` deprecated in favour of `SpaceUtil.Size getRight()`
  - `int getSouth()` deprecated in favour of `SpaceUtil.Size getBottom()`
  - `int getWest()` deprecated in favour of `SpaceUtil.Size getLeft()`  

- `BorderLayout`, `ColumnLayout`, `GridLayout`:
  - All current `int`-based constructors are deprecated and use helpers in `SpaceUtil` to convert int spacing to the enum. The values used for the conversion points are as per the former conversion client UI so there will be **no visible change** in gaps.
  - `int getHgap()` deprecated in favour of `SpaceUtil.Size getHorizontalGap()`
  - `int getVgap()` deprecated in favour of `SpaceUtil.Size getVerticalGap()`

- `FlowLayout`, `ListLayout`, `WList`, `WRow`:
  - All current `int`-based constructors are deprecated and use helpers in `SpaceUtil` to convert int spacing to the enum. The values used for the conversion points are as per the former conversion client UI so there will be **no visible change** in gaps.
  - `int getGap()` deprecated in favour of `SpaceUtil.Size getSpace()`

## Enhancements

- `WAbbrTextRender` changed from XML to HTML with change in UI for #639 as a performance improvement.
- `WDataTableRenderer` no longer renders any internal disabled states with no UI change as these were ignored in the UI.
- `WFieldErrorIndicatorRenderer` will not render with no change, previously ignored in UI.
- `WLabelRenderer` is aware of the nature and state of its labelled component and is able to correctly set the state of a `WLabel` in the UI without cross-reference to its component. This was required to fix several possible errors caused when a `WLabel` was in an AJAX response.
- `WMenuItemRenderer` and `WSubMenuRenderer` will only output the `accessKey` member if the component is at the top level of its `WMenu`
- `WTabGroupRenderer` will no longer paint `WTabGroup` as `ui:tabgroup` but pass through to its `WTabs` with no UI change as previously WTabGroups were ignored.
- Selenium test elements improved and extended.

# JavaScript

- Fixed imports between the `wc/dom` and `wc/ui` namespaces so that no dom-level modules require anything in the `wc/ui` space.
- Changed the require config `baseUrl` to make it shorter as this makes adding application level modules easier. The new BaseUrl only uses path settings which are also available in Java.
- Added new module `wc/ui/getForm` as we have several areas where we need to get the form from a particular element.
- Moved a11y errors for unlabelled components and missing headings/legends from XSLT to debug mode JavaScript.
- Moved determination of position for `WLabel` from XSLT to JavaScript.
- Moved positioning of `WValidationErrors` in-line messages from XSLT to JavaScript.
- Moved positioning of `WSuggestions` from XSLT to JavaScript.
- Moved `submitOnChange` warning from XSLT to JavaScript.
- Moved all AJAX trigger calculations, including a11y attributes, from XSLT to JavaScript.
- Moved aria-live determination from XSLT to JavaScript.
- Simplified row expansion, selection, select toggle (inc row selection) and collapsible toggle (inc row explansion) JavaScript modules and moved control calculation from XSLT to JavaScript.
- Calculation of menu separator orientation moved from XSLT to JavaScript.
- Moved some implicit, non-circular dependencies to explicit dependencies to help clean-up and hunt for circular dependencies.
- Removed the circular dependency between `formUpdateManager` and `cancelUpdate`.
- Improved re-use of Widgets in all components which defined their own fieldset Widget **and** depended on fieldset (implicit or explicit).
- Fixed a bug which could cause a (caught) exception when a WTree's item was selected if the WTree has client expansion and was not an ajax trigger.

# XSLT

- Removed all keys and cross-element look ups.
- Removed all superfluous templates.
- Fixed ambiguous, missing and multiple imports.
- Removed all ancestor lookups for components where the ancestor is **not** guaranteed to be in the same response if the component is inserted as part of an AJAX response.
- Simplified calculations for `ui:label`, `ui:submenu` and `ui:menuitem` by moving some render logic to the Java renderers.
- Moved calculation of margins and size classes by converting the components to use a consistent enum removing the conversion in XSLT.
- All simple inputs are now wrapped in a span element to make them all behave the same.
- Simplified WButton and WLink popup calculations.
- Moved popup specs from its own template to `ui:popup` (where it belonged).

@jonathanaustin @ricksbrown PTAL.


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WMenuItemRenderer.java
```diff
@@ -1,20 +1,57 @@
...
  */
 final class WMenuItemRenderer extends AbstractWebXmlRenderer {
 
 	/**
+	 * A WMenuItem is selectable if it is explicitly set as selectable (or not selectable) or if its own selectability is not set if its nearest
+	 * neighbour selection container is selectable.
+	 * @param item the WMenuItem to test
+	 * @return {@code true} if the WMenuItem is selectable
+	 */
+	private boolean isSelectable(final WMenuItem item) {
```
jonathanaustin: Do you think it is better for this logic to be in WMenuItem.

There is already a "isSelectable" on WMenuItem which is deprecated and just returns "getSelectablility". 

What if the depraction was removed and repurpose the method to include this logic and remove setSelectable.

Might be good to also put the "getRole" logic on WMenuItem.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/XMLUtil.java
```diff
@@ -24,18 +24,7 @@
 	/**
 	 * XML Declaration.
 	 */
-	public static final String XML_DECLERATION = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>";
```
jonathanaustin: This is technically an API change.

Might need to keep XML_DECLERATION and mark as deprecated.

wcomponents-core/src/main/resources/schema/ui/v1/all.xsd
```diff
@@ -1,13 +1,11 @@
...
 	<!--
-	A collection of all the top-level elements in the schema.
-	This is imported by schema.xsd.
+		A collection of all the top-level elements in the schema.
+		This is imported by schema.xsd.
 	-->
-	<xs:include schemaLocation="abbr.xsd" />
 	<xs:include schemaLocation="additionalParameters.xsd" />
 	<xs:include schemaLocation="ajaxTrigger.xsd" />
 	<xs:include schemaLocation="application.xsd" />
```
jonathanaustin: Can "image.xsd" also be removed?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WMenuItemRenderer.java
marksreeves: I can see the point of moving the `isSelected` logic to WMenuItem - even though it is a conversion from earlier client-only code. I do not believe we should do the same with `getRole` as this is **only** client code and is, in the first instance, an indication of the multi-selectability of the ancestor MenuContainer. We could output a selection mode attribute but I chose to output the role attribute as it will become useful when we move away from XSLT.

wcomponents-core/src/main/resources/schema/ui/v1/all.xsd
marksreeves: yes

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/XMLUtil.java
marksreeves: Good point.

PR #1085, opened 2017-02-10
Memory leak detection

ricksbrown:
@jonathanaustin PTAL


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractInput.java
```diff
@@ -38,6 +39,8 @@ public void addValidator(final FieldValidator validator) {
 
 		validator.setInputField(this);
 		model.validators.add(validator);
+		// possible source of memory leaks
```
jonathanaustin: What do you think of calling the method `checkSize`. This indicates what we are `checking`. We dont have to describe the action in the method name as we may change from `logging` to `throw exception`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/MemoryUtil.java
```diff
@@ -0,0 +1,73 @@
+package com.github.bordertech.wcomponents.util;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * This class is intended to help detect and trace possible memory leaks.
+ * The goals are:
```
jonathanaustin: Do you think the class level description should be more generic and the list size description belongs on the `checkSize` method. This allows other checks to be added to the class and their specific test described in their java doc.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/MemoryUtil.java
```diff
@@ -0,0 +1,73 @@
...
+ * - Use a common logger so that these messages can be easily silenced for all classes.
+ * - Provide a stack trace so that possible memory leaks are easier to track.
+ *
+ * @author Rick Brown
+ */
+public final class MemoryUtil {
+
+	/**
+	 * For convenience this can be used as a "rule of thumb" when testing the size of a List of components in the UI.
+	 */
```
jonathanaustin: What if this was a runtime parameter that defaults to 1000. This would allow projects to adjust the value.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractInput.java
ricksbrown: Yes good call.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/MemoryUtil.java
ricksbrown: OK

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/MemoryUtil.java
ricksbrown: Yes indeed, it was on my todo list but I forgot. Good call.

Issue #605, opened 2016-05-10
WComponents Selenium Framework


Joshua-Barclay:
Selenium WebDriver works out of the box with WComponents (e.g. find by ID, tag, xpath), however some elements require intimate knowledge of the current client-side implementation. e.g. getting the title of a WDialog, or the pagination buttons for a table.

Ideally apps should be able to write tests that are unaware of the implementation details.
The best way we can do this is by writing a WComponent Selenium testing framework.

I am considering two possible implementations

**Option 1 - Wrap WebDriver**
This is the heavyweight option, but it is how Vaadin has solved the same problem.
We could create an implementation of Selenium WebDriver for WComponents that wraps the actual driver implementation. The wrapped driver could wrap WebElements as WCompWebElements and would provide an additional API to simplify access:
`WCompSeleniumTable table = driver.getTable(id);`
`WCompWebElement e = driver.getElement(id);`
**Pros:** Can automatically include the 'wait' code where needed. Intuitive for apps to use, requires almost no additional code. Can overload all 'find' methods to take WComponents as a parameter.
**Cons:** Highly coupled to Selenium implementation, more significant development effort.

**Option 2 - Static Utility Class**
The lightweight approach is to create a static utility class which will provide convenience methods:
`WCompSeleniumTable table = WCompSelenium.getTable(driver, id);`
`WCompSelenium.waitForPageReady(driver, timeout);`
**Pros:** low coupling with Selenium implementation, better suited to ad-hoc usage.
**Cons:** Less intuitive for apps, requires a lot of additional code

At the moment I am strongly leaning towards option 1. It will require more development effort but I believe it will be a better result for applications to use. We will also need to keep a close eye on Selenium updates to ensure the WComponent driver continues to support the full API.

@jonathanaustin @JohnMcGuinness @ricksbrown @marksreeves Comments welcome please.

JohnMcGuinness:
I would definitely support option 1.

I disagree with Option 2 Pro regarding ad-hoc usage. I think that Option 1 when fully implemented would be easier to use for ad-hoc testing than option 1.

@jonathanaustin @ricksbrown @marksreeves @Joshua-Barclay

ricksbrown:
Would it be feasible to do both like so:

1. Implement Option 2
2. Implement Option 1 by leveraging Option 2

In other words we write loosely coupled static utility classes but do not expect anyone else to use them but ourselves. Then we add the layer that is heavily coupled to selenium.
[...]

PR #1089, opened 2017-02-14
Add Selenium helper for WLabel

marksreeves:
Add new WebElement extension for WLabel and two methods for getting a WLabel by its text content, one with an index for getting the nth of several labels with the same text content.


marksreeves:
Part of #605

PR #1090, opened 2017-02-14
Client debug property

### LINKED ISSUE #1012: Client debug mode should be tied to java debug mode
We have two configuration properties to set debug mode:

- `bordertech.wcomponents.debug.enabled` and 
- `bordertech.wcomponents.debug.clientSide.enabled`.

To enable client side debugging both of these must be set true. This is completely superfluous and client debuggon should be on by default when WComponents debug mode is on.

marksreeves:
Ensure that debug mode is consistent for server and client code. Found
and fixed a HTML error in DevToolkit_header.vm.

Resolves #1012.

@jonathanaustin PTAL


wcomponents-lde/src/main/java/com/github/bordertech/wcomponents/lde/PlainLauncher.java
```diff
@@ -109,8 +103,6 @@ public synchronized WComponent getUI(final Object httpServletRequest) {
 	@Override
```
jonathanaustin: This method could be removed as it only calls super.

wcomponents-lde/src/main/java/com/github/bordertech/wcomponents/lde/PlainLauncher.java
```diff
@@ -124,19 +116,7 @@ public InterceptorComponent createInterceptorChain(final Object request) {
 		// The toolkit must render itself within the page shell
 		// and wrap the main WComponent output.
 		InterceptorComponent.replaceInterceptor(PageShellInterceptor.class,
```
jonathanaustin: As this is now replacing itself with itself, this override method is no longer needed.

wcomponents-lde/src/main/java/com/github/bordertech/wcomponents/lde/PlainLauncher.java
marksreeves: 👍

wcomponents-lde/src/main/java/com/github/bordertech/wcomponents/lde/PlainLauncher.java
marksreeves: 👍

marksreeves:
@jonathanaustin all items in review completed. Thanks.

PR #1112, opened 2017-03-01
Autocomplete settings

### LINKED ISSUE #1007: Allow for autofill/autocomplete to be switched for supporting components
Components which produce form elements which support the autocomplete attribute should allow this attribute to be set/unset. The default is to not set the autocomplete attribute.

**note:** @jonathanaustin 

As a bare minimum we should allow autocomplete to be turned off by having a boolean isAutoComplete() where the default is `true`. This is the worst option.

If we want to do this properly then the autocomplete attribute should allow for [autofill detail tokens](https://html.spec.whatwg.org/multipage/forms.html#autofill-detail-tokens) and not just `on`/`off`. One step up from the bare minimum is to allow tokens to be set from an enum and ignore `section-*` values. The best option is to allow any valid value and make autocomplete a String member (this should be our default best option for any new attribute we have to expose as it reduces the abstraction of the abstraction layer).

The components which are affected are:

- WDateField
- WDropdown (see note below)
- WEmailField (see note below)
- WMultiDropdown
- WMultiSelect
- WMultiTextField
- WPasswordField
- WPhoneNumberField (see note below)
- WNumberField
- WSingleSelect
- WTextField (see note below)

## Notes

1. autocomplete must be `off` (and therefore any other option[s] will not be supported in the client) for:

    - any WTextField, WEmailField or WPhoneNumberField which is associated with a WSuggestions;
    - WPartialDateField; and
    - WDropdown in the `COMBO` state.

2. Autocomplete causes some obscure bugs in IE (8 and before, possibly later) and so it is deemed incompatible with that browser. If an implementation needs IE support we cannot (necessarily) provide autocomplete support.

3. I very much doubt that autocomplete could be sensibly or appropriately used on WMultiTextField or WMultiDropdown so these _could_ be a binary `on`/`off`where `on` is the default and not output in the XML.

4. autocomplete is not useful for WShuffler or WMultiSelectPair and will be set to 'off' in the select elements in these components.

marksreeves:
Allow setting of input component’s autocomplete settings (where appropriate and supported). Resolves #1007.


marksreeves:
Decided to implement in a different way. Closing pull request without action.

PR #1141, opened 2017-03-25
Refactor SpaceUtil.Size to Size

### LINKED ISSUE #1131: Suggestions for improving usage of Size
Currently the ```Size``` enum is nested within the ```SpaceUtil``` class. Other than the ```Size``` enum the ```SpaceUtil``` class has utility methods that allow for converting between ```Size``` and ```int```. These methods are deprecated as is everything else in the class except the ```Size``` enum. Once the deprecated members of this class are removed, the API will have an enum wrapped in a class with no other members in the class.

I would like to propose that the ```Size``` enum be removed from the ```SpaceUtil``` class so that it is defined in its own file, and so ```SpaceUtil``` is not then a part of any public API.

I'd be happy to submit a PR if you all agree with these suggestions.

WDYT @jonathanaustin @marksreeves @Joshua-Barclay

JohnMcGuinness:
Should resolve #1131


marksreeves:
Thanks for that John - looks good. I am going to defer the merge to @jonathanaustin

jonathanaustin:
LGTM

PR #1179, opened 2017-04-26
prevent double handling of images

### LINKED ISSUE #1158: Client handlebars causes double request for images
The way i18n is done in-client causes images to be requested twice. Not a big deal if images are cached but annoying and potentially costly for uncached images.

The current workaround is to turn on server rendering of handlebars. We should be able to do better than this.

To use the workaround set a property:

```
bordertech.wcomponents.template.render.mode=on
```

### Note:
Turns out this does not work in 1.3 as we do not test for server handlebars.

marksreeves:
Fix #1158

* Port URL encoding and brace handling updates from eddie.


marksreeves:
@jonathanaustin the Java changes are mostly but not entirely from cherry picking your fixes. PTAL.

Issue #1138, opened 2017-03-23
Improve performance of wcomponents-test-lib


marksreeves:
These tests seem to take far longer than they ought. Much of this may be due to some slowness in `com.github.bordertech.wcomponents.test.selenium.ByLabel` which is a complex XPath locator and has already been identified having poor performance. 

Also check the other By extensions for similar issues. ByWComponent, for example, has an xpath locator to find by ID which seems a little redundant.

PR #1193, opened 2017-05-08
Selenium Performance

jonathanaustin:
Improve Selenium tests and associated helpers.

Related to #1138.

The methods in `TreeUtil` (i.e. `findWComponent` and `findWComponents`) that use a `path` to find components has been changed to search only visible components. New `find` methods that provide a `visible only` boolean flag have been included to toggle this. This change will effect selenium tests that use `byWComponentPath` with invisible components in their path. Either change the path to use only `visible` components or use the new `constructor` on `byWComponentPath` that allows the visible only flag to be toggled.

Fixed issue in `ByWComponentPath` that was leaving the `user context` on the Thread. Unit tests that check a component's state may have depended on this wrong behaviour. These tests will need to be changed to `push` and `pop` the `User Context` when checking the components state. `SeleniumWComponentTestCase`  provides a new helper method `getUserContextSession()` to retrieve the `user context`.

```java
UIContext uic = getUserContextForSession();
UIContextHolder.pushContext(uic);
try {
// Check component state
} finally {
    UIContextHolder.popContext();
}
```
`SeleniumWComponentsWebDriver` and `SeleniumWComponentWebElement` provide new `findImmediate` methods that can be used to `find` an element without the `wait implicit`. This assumes the page is already loaded and will return immediately if the element is not present. They also both provide a new helper method `getUserContextSession()` to retrieve the `user context`.


marksreeves:
Looks nice.

PR #1201, opened 2017-05-11
Selenium Performance

jonathanaustin:
Wrap example in WApplication.

Related to #1138.


jonathanaustin:
Checks are marked as failed as they were rerun but failed due to the change already being merged. They did originally pass.

PR #1202, opened 2017-05-11
Selenium Performance

jonathanaustin:
Move ServerStartStopListener into test-lib.
Travis and Circle CI now have distinct build and test phases.
Allow WComponent Selenium Tests to run in parallel or in series.
Copy I18N resources for WComponent builds.

Relates to #1138.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-before_install: echo "MAVEN_OPTS='-Xmx512m -XX:MaxPermSize=128m'" > ~/.mavenrc
+before_install:
+- echo "MAVEN_OPTS='-Xmx512m -XX:MaxPermSize=128m'" > ~/.mavenrc
+- mvn clean
 
 language: Java
 jdk:
@@ -10,7 +12,6 @@ env:
    #   via the "travis encrypt" command using the project repo's public key
    - secure: "MAirt38nl74gOmupg6oWoghoXT8YDgjMJmPzrv6YcMLJ92tS8kv7PGtSfdM+BlxXwlvvXIlqXNwb2R+U4CnjsZUsL+28RGqFiKReAbDUON55Cucb/Yy0ZB+QC9NpCxM5LfC4BycDPI6HStx7ZmDy61/U9alogrbhiqomD/LlMTQ="
 
-before_script:
- - mvn clean install -Dmaven.javadoc.skip=true -DskipTests
+## Travis installs the project with the following maven command:- "mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V"
 script:
  - mvn test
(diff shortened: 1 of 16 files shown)


jonathanaustin:
@ricksbrown Can you please check how I am "building" the I18N details manually for the "theme". Instead of "unpacking" the details from the "jar". Maven is complaining about the "jar" not being packed in the "test" phase.

wcomponents-theme-parent/pom.xml
```diff
@@ -125,7 +128,7 @@
 							<goal>unpack</goal>
 						</goals>
 						<configuration>
-							<skip>${theme.skip.antrun}</skip>
+							<skip>${theme.skip.unpack}</skip>
```
jonathanaustin: @ricksbrown Changed the unpack of i18n to be used by projects that build their own theme. So instead of "skipping" on "antrun" it now skips on "unpack".

wcomponents-theme-parent/pom.xml
```diff
@@ -265,6 +268,27 @@
 							<goal>run</goal>
 						</goals>
 					</execution>
+					<!-- Build the i18n Resources from the wcomponents-i18n module. -->
```
jonathanaustin: @ricksbrown This is where the "manual" build of I18N is done. This duplicates the build in wcomponents-i18n module. Thsi gets around the issue of the wcompoentns-i18n" jar not being packed before the test phase. This is not an issue for projects that consume wcomponents as wcomponents-i18n is already in the repo.

wcomponents-theme/unbuilt.package.json
```diff
@@ -18,6 +18,8 @@
 		"handlebars": "~4.0.5",
 		"i18next": "~3.4.1",
 		"i18next-xhr-backend": "~1.0.1",
```
jonathanaustin: Override the versions brought in by intern for "digdug" and "leadfoot" that is using "~". The latest version of these breaks saucelabs.

marksreeves: This is a bit sub-optimal as it may result in us breaking Intern at some point. We cannot comment a JSON file (because JSON is rubbish) but we need to put a note somewhere for whoever upgrades Intern in the future. I do not have a good solution to this.

jonathanaustin: What if we put a comment in the theme-parent pom.xml above the intern version parameter.

marksreeves: I was thinking along similar but more extreme lines: make the two packages into a property defined in the POM. @ricksbrown thoughts on this?

.travis.yml
```diff
@@ -10,14 +12,6 @@ env:
...
-    project:
-      name: "BorderTech/wcomponents"
-      description: "Build submitted via Travis CI"
-    notification_email: jonoa4n@gmail.com
-    build_command_prepend: "mvn clean"
-    build_command: "mvn -DskipTests compile"
-    branch_pattern: coverity_scan
-
-script: "mvn verify"
+## Travis installs the project with the following maven command:- "mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V"
```
ricksbrown: This comment is certainly essential. Until I read it I was wondering where the `install` phase was.

wcomponents-theme-parent/pom.xml
ricksbrown: I think this makes sense, nice work ✨

PR #1211, opened 2017-05-24
QC165120 encoded url components are being decoded

ricksbrown:
This is basically reverting to previous behaviour.
This needs to be thought out in detail before we start messing with URLs.
There is one unit test, my hunch is it needs at least 10 to cover the various permutations.

Before this change this `http://www.example.com/&foo=fu%2Fbar` would be rendered as this `http://www.example.com/&foo=fu/bar` and that's not right at all.

My gut feeling is that URL encoding (all or part of the URL) is the responsibility of the application developer. If we give it some thought maybe there is a way.... I guess if there's a space in the URL it hasn't been encoded (but maybe part of it has?)


marksreeves:
Sorry @ricksbrown I am closing this PR and re-doing it. Code change is good but test utility wasn't updated and you aren't available to grab your branch. It won't take long.

Issue #1255, opened 2017-06-14
AbstractInput submitOnChange is a source of a11y failure


marksreeves:
We currently output **visible** a warning if `submitOnChange` is set. This is sufficient to prevent an _unexpected_ change of context and is usually enough to stop people using it. It is, however, time we dropped `submitOnChange` - it serves no useful purpose and should be replaced with a WAjaxControl and/or better design. 

This predates all of us I think and should have never been built.

See https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html

PR #1259, opened 2017-06-14
Bug fixes

### LINKED ISSUE #501: Automatic re-focus breaks a11y
A number of components call setFocussed() to automatically re-focus themselves when a view is rendered. For example a WButton will regain focus if it is in the view which is loaded after it is used to invoke a https POST "round trip". This was originally introduced at the behest of an accessibility consultant who appears to have not been aware of the meaning of SC 2.4.3

This **must** be removed as it is by no means guaranteed that the control should retain focus. Under **most** circumstances only one of the following two scenarios are valid after a "round trip":
1. the same view is displayed with validation errors, in which case the validation error list should have default focus; or
2. a new view is displayed, in which case focus **must not** be set to the old button.
## Example

A multi-view application has a WButton used to save each view in a sequential process. Each step of the process is a unique view but the "Save" button is shared. It would be completely inappropriate for this button to retain focus after causing the sequence to proceed from step `N` to step `N + 1`.
## Proposed solution

Remove all in-built calls to `setFocussed()` but retain `setFocussed()` so that each use-case may be ascertained on its merits thereby moving the determination of appropriate focus retention to the only place it can be adequately specified: the consuming application.

### LINKED ISSUE #1253: Unexpected focus/scroll when error present
If 
1. a WMessages or WMessageBox is used to display a page level error; and
2. the error is not focusable (so not generated from WValidationErrors); and
3. the page loads with a default focus set (e.g. from WButton auto-refocus as per #501 )
then the button will receive focus and the page may scroll to hide the messages.

We should check for error boxes before doing autofocus. We _could_ check for all boxed messages.

@ricksbrown I think it should be enough to not honour focus request (on page load - not from an Ajax response) if there is an **ERROR** level message box. We could test document position to try to determine if the element to be focussed is above the message box but I think this is overkill. If there are new error messages we should assume a change of context and not set focus within the page content. Thoughts?

### LINKED ISSUE #1257: Compound components may not get a legend
When a compound control which outputs a fieldset (`WCheckBoxSelect` etc) is is rendered in XML it does not know about its `WLabel`. We have a JS module which gathers all these fieldsets, finds their faux-labels (the output of `WLabel`) and uses those faux-labels to create a legend for the fieldset.

If the component has a title (`setToolTip(String)`) or aria-label (`setAccessibleText(String)`) then a `WLabel` is not _needed_ but a legend **is**. In these cases the legend should be generated from the attribute and the attribute should be removed (note there is CSS to hide the faux label and show the legend EXCEPT in `WFieldLayout` with layout `LAYOUT_FLAT` and in that case the legend is hidden instead - I _really_ don't like WComponents some times!).

Anyway, when the component does not have a `WLabel` is does not currently get a legend and it should.

### LINKED ISSUE #1258: WRadioButton.handleRequest can throw a null pointer exception
The test 
``` java
if (changed && isSubmitOnChange() && UIContextHolder.getCurrent().getFocussed() == null) {
  setFocussed();
}
```

Can throw a null pointer exception if `UIContextHolder.getCurrent() == null`. This exact scenario is present in the unit tests but the lack of a current context is masked by the test of `sSubmitOnChange` (which is not itself bound to there being a current UIContext).

@jonathanaustin comments?

marksreeves:
Fix #1258
Fix #1257
Fix #1253
Resolve #501
Address part of #1255

Update eslint and fix a number of JavaScript issues.

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractInput.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractInput.java
@@ -296,24 +296,25 @@ public abstract class AbstractInput extends WBeanComponent implements Input {
 
 	/**
 	 * Perform change logic for this component.
+	 * <p>Reset focus ONLY if the current Request is an Ajax request. See https://github.com/BorderTech/wcomponents/issues/501.</p>
 	 */
 	protected void doHandleChanged() {
 		// If there is an associated action, execute it
 		if (getActionOnChange() != null) {
 			final ActionEvent event = new ActionEvent(this, getActionCommand(), getActionObject());
+			final WComponent currentInput = this;
 			Runnable later = new Runnable() {
 				@Override
 				public void run() {
 					getActionOnChange().execute(event);
-
-					if (isSubmitOnChange() && UIContextHolder.getCurrent().getFocussed() == null) {
+					if (AjaxHelper.isCurrentAjaxTrigger(currentInput) && UIContextHolder.getCurrent().getFocussed() == null) {
 						setFocussed();
 					}
 				}
 			};
 
 			invokeLater(later);
-		} else if (isSubmitOnChange() && UIContextHolder.getCurrent().getFocussed() == null) {
+		} else if (AjaxHelper.isCurrentAjaxTrigger(this) && UIContextHolder.getCurrent().getFocussed() == null) {
 			setFocussed();
 		}
 	}
@@ -382,7 +383,10 @@ public abstract class AbstractInput extends WBeanComponent implements Input {
 	 * Setting this flag to true will cause this component to post the form to the server when it changes.
 	 *
 	 * @param flag if true, the form is submitted when the component changes.
+	 * @deprecated 1.4.0 as it results in a level A accessibility problem See
+	 * https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html.
 	 */
+	@Deprecated
 	void setSubmitOnChange(final boolean flag) {
 		setFlag(ComponentModel.SUBMIT_ON_CHANGE_FLAG, flag);
 	}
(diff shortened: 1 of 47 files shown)


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractInput.java
```diff
@@ -296,24 +296,25 @@ protected boolean beforeHandleRequest(final Request request) {
...
 		if (getActionOnChange() != null) {
 			final ActionEvent event = new ActionEvent(this, getActionCommand(), getActionObject());
+			final WComponent currentInput = this;
 			Runnable later = new Runnable() {
 				@Override
 				public void run() {
 					getActionOnChange().execute(event);
-
-					if (isSubmitOnChange() && UIContextHolder.getCurrent().getFocussed() == null) {
+					if (AjaxHelper.isCurrentAjaxTrigger(currentInput) && UIContextHolder.getCurrent().getFocussed() == null) {
```
jonathanaustin: What might be helpful is a new method in `AbstractInput` to check if the input is the current AJAX trigger. This could be useful for many input components that do this check....
```
     /**
	 *
	 * @return true if this input is the current AJAX trigger
	 */
	public boolean isCurrentAjaxTrigger() {
		return AjaxHelper.isCurrentAjaxTrigger(this);
	}
```

Also noticed `WTab` has a static method `isCurrentAjaxTrigger`. Might be good to deprecate this and call the AjaxHelper method.

jonathanaustin: What might be helpful is a new method in `AbstractInput` to check if the input is the current AJAX trigger. This could be useful for many input components that do this check....
```
     /**
	 *
	 * @return true if this input is the current AJAX trigger
	 */
	public boolean isCurrentAjaxTrigger() {
		return AjaxHelper.isCurrentAjaxTrigger(this);
	}
```

Also noticed `WTab` has a static method `isCurrentAjaxTrigger`. Might be good to deprecate this and call the AjaxHelper method.

PR #1276, opened 2017-07-26
bring master up to last 1.3 release

marksreeves:
`master` was last updated to 1.3.3 but `eddie` (the last complete stream) peaked at 1.3.13.


ricksbrown:
Eeek, don't merge this before we talk please.

ricksbrown:
`master` is currently pointing to the common ancestor of eddie (1.3) and freddie (1.4).

Any work that happened after that is likely completed manually in both branches (or deliberately excluded). Either way, if we go beyond the common ancestor the merge hell that will ensure when we try to merge in 1.4 will essentially mean that `master` becomes its own unique branch with its own commits that reflects nothing ever deployed to prod or tested ever and is probably broken due to really complicated merges gone wrong.

I reckon we wait until 1.5 is created and then update `master` with the common ancestor of 1.4 and 1.5.

The command `git merge-base` is our friend here and is how I updated `master` to the commit it currently points to.

marksreeves:
DO NOT MERGE

PR #1306, opened 2017-08-23
Default tab mode, deprecations

marksreeves:
* Added methods to WTabSet to allow easier/less verbose creation of tabs.
  - updated calls to WTab constructors to ensure the only constructor called from within WTabSet is the one with all args. This should allow us to remove the other package protected constructors
* Deprecated the unused members
* Deprecated action on change as its application is inconsistent and these actions are conceptually wrong: an action on tab change is a side-effect.

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTabSet.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTabSet.java
@@ -212,7 +212,7 @@ public class WTabSet extends AbstractNamingContextContainer implements Disableab
 	 * @return the tab which was added to the tab set.
 	 */
 	public WTab addTab(final WComponent content, final String tabName) {
-		WTab tab = new WTab(content, tabName, TabMode.LAZY);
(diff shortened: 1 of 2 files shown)


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTabSet.java
```diff
@@ -281,11 +314,22 @@ public void addSeparator() {
...
+		addTab(tab);
+	}
+
+	/**
+	 * Adds a {@link WTab} to the tab set.
+	 *
+	 * @param tab the tab to add.
+	 * @return the tab which was added to the tab set.
+	 */
+	private WTab addTab(final WTab tab) {
```
jonathanaustin: As the Constructor for `WTab` is `protected`, the `addTab` method should probably be `protected` as well, which would allow the class to be extended. 
What do you think?

marksreeves: Sure will change

PR #1341, opened 2017-09-15
Added JavaScript unit tests for module `wc/ui/label`.

marksreeves:
* Found a flaw in label conversion when inputs are not wrapped
* Found a flaw in module `wc/dom/shed` (yes really):  `isMandatory` was out of step with `isHidden` and `isDisabled` in that it was using getAttrbute(`required`) rather than `hasAttribute("required")`. This leads to false negatives when the mark-up is HTML syntax rather than XML syntax. Added another shed unit test for this condition.

Added new unit tests for modules `wc/Observer`.and `wc/xml/xmlString` just to cover some missing lines.


marksreeves:
@ricksbrown including you as a reviewer because I have changed shed.

PR #1350, opened 2017-09-22
Sync uploads

ricksbrown:


ricksbrown:
Added then remove synchronous file uploads, reverted the core change but kept a lot of the refactoring and rationalisation.

PR #1373, opened 2017-10-05
File type and size validation

kandula-aswin:


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -79,6 +86,14 @@ public void setFileTypes(final String[] types) {
 	public void setFileTypes(final List<String> types) {
 		getOrCreateComponentModel().fileTypes = types;
 	}
+	
+	/**
+	 * Checks if one or more file type is supplied.
+	 * @return True/False
+	 */
+	public boolean isSetFileTypes() {
```
jonathanaustin: what about this method being called "hasFileTypes"

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -98,6 +113,14 @@ public void setMaxFileSize(final long bytes) {
 	public long getMaxFileSize() {
 		return getComponentModel().maxFileSize;
 	}
+	
+	/**
+	 * Checks if max file size is supplied.
+	 * @return True/False
+	 */
+	public boolean isSetFileSize() {
```
jonathanaustin: Maybe same with this method "hasMaxFileSize"

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -110,13 +133,109 @@ protected boolean doHandleRequest(final Request request) {
...
+				addFileSizeValidation(validFileSize, InternalMessages.DEFAULT_VALIDATION_ERROR_FILE_WRONG_SIZE, 
+						Util.readableFileSize(value.getSize()), Util.readableFileSize(getMaxFileSize()));
+			} else {
+				validFileSize = true;
+			}
+			
+			// if file is valid, the update data
+			if (validFileSize && validFileType) {
+				setData(value);
+			} else {
```
jonathanaustin: If there was a file previously loaded, this will need to be cleared (ie set null) and return "true" as changed.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -110,13 +133,109 @@ protected boolean doHandleRequest(final Request request) {
...
+	/**
+	 * Indicates whether the uploaded file is valid, if {@link getFileTypes()} is set.
+	 *
+	 * @return true if file type valid, false file type invalid, otherwise null.
+	 * @see {@link getFileValidationMessages()}
+	 */
+	public Boolean isFileTypeValid() {
+		return getComponentModel().validFileType;
+	}
+
```
jonathanaustin: What do you think? If no file size is set, then is it a valid file size? Maybe this should return TRUE when no file size is set. Same with File type?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -110,13 +133,109 @@ protected boolean doHandleRequest(final Request request) {
...
+	/**
+	 * Indicates whether the uploaded file is valid, if {@link getMaxFileSize()} is set.
+	 *
+	 * @return true if file size valid, false file size invalid, otherwise null.
+	 * @see {@link getFileValidationMessages()}
+	 */
+	public Boolean isFileSizeValid() {
+		return getComponentModel().validFileSize;
+	}
+
```
jonathanaustin: ValidateComponent will also need to be overridden to include the new validation messages.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
```diff
@@ -670,6 +691,25 @@ protected void doHandleUploadRequest(final Request request) {
 
 		// Wrap the file item
 		FileItemWrap wrap = new FileItemWrap(items[0]);
+		
+		// if fileType is supplied then validate it
+		if (isSetFileTypes()) {
```
jonathanaustin: You could combine these two if statements (I have been pinged by static analysis on this before)

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
```diff
@@ -670,6 +691,25 @@ protected void doHandleUploadRequest(final Request request) {
...
+			if (!FileValidationUtil.validateFileType(wrap, getFileTypes())) {
+				String invalidMessage = String.format(I18nUtilities.format(null, InternalMessages.DEFAULT_VALIDATION_ERROR_FILE_WRONG_TYPE), 
+						StringUtils.join(getFileTypes().toArray(new Object[getFileTypes().size()])));
+				throw new SystemException(invalidMessage);
+			}
+		}
+
+		// if fileSize is supplied then validate it
+		if (isSetFileSize()) {
```
jonathanaustin: Could also combine these if statements

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/InternalMessages.java
```diff
@@ -119,6 +119,16 @@
...
+	
+	/**
+	 * The key for invalid file type.
+	 */
+	public static final String DEFAULT_VALIDATION_ERROR_FILE_WRONG_TYPE = "bordertech.wcomponents.message.fileWrongType";
+
+	/**
+	 * The key for invalid file size.
+	 */
+	public static final String DEFAULT_VALIDATION_ERROR_FILE_WRONG_SIZE = "bordertech.wcomponents.message.fileTooLarge";
```
jonathanaustin: Might be better to match the parameter name and variable name (same as the file wrong type), just for consistency

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/Util.java
```diff
@@ -129,4 +131,19 @@ public static String leftTrim(final String aString) {
 		return aString.substring(start);
 	}
```
jonathanaustin: What if this was included in the FileValidationUtil? Which means maybe FileValidationUtil might need to be named a bit more generic like FileUtil.

ricksbrown:
Should the client and server agree on how many bytes in a KB (for example)? Messages to the user could be confusing. The client currently uses 1000 bytes in a KB (as per IEC 80000-13)

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
kandula-aswin: When fileSize is not set it returns a null, same with fileTypes. What do you think?

marksreeves:
There are 1000 bytes in a kilobyte. It is an SI unit and uses the normal SI size prefix and the correct definition of that prefix. 

Using 2^n as a "close enough and we're geeky" variation on an SI unit is not acceptable.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
jonathanaustin: Returning null is always dangerous. I think it should be `boolean` and return `true` even if no file type set as technically it is valid. Same for `maxfilesize`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -79,6 +88,14 @@ public void setFileTypes(final String[] types) {
 	public void setFileTypes(final List<String> types) {
 		getOrCreateComponentModel().fileTypes = types;
 	}
+	
+	/**
+	 * Checks if one or more file type is supplied.
+	 * @return True/False
+	 */
+	public boolean hasFileTypes() {
```
jonathanaustin: This method name looks a lot better.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -110,13 +135,97 @@ protected boolean doHandleRequest(final Request request) {
...
+		return getComponentModel().validFileSize;
+	}
+	
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	protected void validateComponent(final List<Diagnostic> diags) {
+		super.validateComponent(diags);
+
```
jonathanaustin: This validation logic looks good. 
What do you think of putting the `create error message` in the `FileUtil` so it can be shared between both file widgets.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -110,13 +135,97 @@ protected boolean doHandleRequest(final Request request) {
...
+	}
+	
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	protected void validateComponent(final List<Diagnostic> diags) {
+		super.validateComponent(diags);
+		
+		if (BooleanUtils.isFalse(isFileTypeValid())) {
```
jonathanaustin: If `isFileFileTypeValid` becomes `boolean` then you wont need the `BooleanUtils`

marksreeves:
One more thing about `maxFileSize` (WMultiFileWidget line 914) - as well as correcting the previous coder's erroneous assumption - We do not provide a default `maxFileSize` in WFileWidget (line 244). Should we make this consistent?

kandula-aswin:
@marksreeves the `maxFileSize` validation simply compares the uploaded `fileItem` size against `max size` (in bytes), if it fails then validation message takes 1000 bytes in a kilobyte and so on to construct readable diagnostic message

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
ricksbrown: May I chime in with my opinions? Thanks... (feel free to disagree)

* By default ALL input is valid and accepted - we should be liberal in what we accept (is this the Robustness Principle?). 

  Constraints may be given but if they are not then there are no constraints *from a validation perspective*. There will be natural constraints but that is different to **validity**. 

  For example if we have a text input with no constraints then any input is valid. If we constrain it to only allow 20 characters with no whitespace then it is only valid if it meets both those constraints.

* Booleans should not be used to represent more than two states.
  I believe the API would be more intuitive if we used an explicit three states (e.g. an Enum).
  In this case it is purely academic because we only need two states.

kandula-aswin: Yes agree @ricksbrown my last commit has fixed it, all Boolean are changed to boolean. Awaiting review.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
marksreeves: This is not-picky but the Javadoc is incorrect. A boolean returns `true` or `false` not `True/False` and perhaps the returns snippet should indicate the conditions under which the return is either using the snippet `{@code true}` and/or `{@code false}`.

marksreeves:
@jonathanaustin can you please revisit your review. Are your last two comments part of the changes requested or clarification? If there are no more changes required can you please finalise your review so that @kandula-aswin can update this PR to the current code state to fix the changelog conflict so we can merge it.

kandula-aswin:
I have merged the latest `upstream/freddie`, and resolved the conflict in `CHANGELOG.md`

kandula-aswin:
@jonathanaustin please review, incorporated your suggestions

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -79,6 +84,15 @@ public void setFileTypes(final String[] types) {
 	public void setFileTypes(final List<String> types) {
 		getOrCreateComponentModel().fileTypes = types;
 	}
+	
+	/**
+	 * @see setFileTypes(List<>)
+	 * @see setFileTypes(String[])
+	 * @return {@code} true if one or more file type is supplied.
+	 */
+	public boolean hasFileTypes() {
```
jonathanaustin: A small performance improvement is to call `getComponentModel` once as it can be expensive.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -110,12 +132,86 @@ protected boolean doHandleRequest(final Request request) {
 		boolean changed = value != null || current != null;
```
jonathanaustin: Can `value` be `null`? If not this test can actually be a bit confusing. This is existing code, but would be good to look at and clarify and deal with a `null` value correctly.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -110,12 +132,86 @@ protected boolean doHandleRequest(final Request request) {
...
+			}
+		} 
 
 		return changed;
 	}
 
+	/**
+	 * Reset validation state.
+	 */
+	private void resetValidationState() {
```
jonathanaustin: Again, just retrieve the model once. Another performance improvement is to check if you have to update the model in the first place by checking if the values have changed.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/mock/MockRequest.java
```diff
@@ -97,7 +99,6 @@ public void addParameterForButton(final UIContext uic, final WButton button) {
 	/**
 	 * Sets mock file upload contents.
 	 *
-	 * @deprecated Because it appears unused
```
jonathanaustin: We cant deprecate this based on our usage. As it is public it maybe used by projects using `WComponents`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
```diff
@@ -0,0 +1,108 @@
+package com.github.bordertech.wcomponents.util;
```
jonathanaustin: Some unit tests specific for FileUtil would be good.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
```diff
@@ -273,6 +275,15 @@ private boolean isValid(final String fileType) {
 		result = new ArrayList<>(fileTypes);
 		return result;
 	}
+	
+	/**
+	 * @see setFileTypes(Collection<>)
+	 * @see setFileTypes(String[])
+	 * @return {@code} true if one or more file type is supplied.
```
marksreeves: rewrite this as follows: 
```
@return {@code true} if one or more file type is supplied.
```

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
kandula-aswin: Below is logic 
```java
boolean changed = value != null || current != null;
```
Test `WFileWidget_Test.testDoHandleRequest`, following situations are covered
1) value != null and current = null -> changed = true
2) value != null and current != null -> changed = true
3) value = null and current != null -> changed = true
4) value = null and current = null -> changed = false

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/mock/MockRequest.java
kandula-aswin: Originally this method had a `@deprecated` in it's javadoc, I have removed it.

jonathanaustin: ooops...You deleted it. Good job.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -166,8 +167,13 @@ protected boolean doHandleRequest(final Request request) {
 	 * Reset validation state.
 	 */
 	private void resetValidationState() {
-		getOrCreateComponentModel().validFileType = true;
-		getOrCreateComponentModel().validFileSize = true;
+		// Update the model only if the original state has changed
+		if (!getComponentModel().validFileSize) {
+		    getOrCreateComponentModel().validFileType = true;
```
marksreeves: Why are you testing `validFileSize` before setting `validFileType`? Are you ure this is correct?

wcomponents-i18n/src/main/resources/com/github/bordertech/wcomponents/messages.properties
```diff
@@ -39,6 +39,8 @@ bordertech.wcomponents.message.fieldDateAfterToday={0} must be after today.
 bordertech.wcomponents.message.fieldMinSelect={0} must have {1} or more options selected.
 bordertech.wcomponents.message.fieldMaxSelect={0} must have {1} or less options selected.
 bordertech.wcomponents.message.fieldInvalidPattern={0} is not in the required format.
+bordertech.wcomponents.message.fileWrongType=The file you have selected is not of an accepted type. Only the following type/s are accepted: %s.
+bordertech.wcomponents.message.fileWrongSize=The file you have selected is too large. Maximum file size is %s.
```
kandula-aswin: @marksreeves what do you think of these validation messages. Are they fit for purpose, or too long?

marksreeves: They are equivalent to the client-side messages. Length is not really a major problem in this case: better to be long and clear than brief and ambiguous or misleading.

kandula-aswin: Thanks @marksreeves yes you are correct, I did copy the text from https://github.com/BorderTech/wcomponents/blob/5eccfb58835ea250196f8b6cccde802bd76df686/wcomponents-i18n/src/main/resources/com/github/bordertech/wcomponents/theme-messages.properties#L292

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -167,11 +167,10 @@ protected boolean doHandleRequest(final Request request) {
 	 * Reset validation state.
 	 */
 	private void resetValidationState() {
-		// Update the model only if the original state has changed
-		if (!getComponentModel().validFileSize) {
+		// Update the model only if the original state has changed. A user specific model is created
+		// if state changes for either of the fields.
+		if (!getComponentModel().validFileSize || !getComponentModel().validFileSize) {
```
marksreeves: Please read line 172 carefully and consider the latest change you have made.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -135,17 +135,26 @@ protected boolean doHandleRequest(final Request request) {
 		if (changed) {
 			// Reset validation fields
 			resetValidationState();
-			
+			// if User Model exists it will be returned, othewise Shared Model is returned
+			final FileWidgetModel sharedModel = getComponentModel();
+			// if User Model exists it will be returned, othewise it will be created
+			final FileWidgetModel userModel = getOrCreateComponentModel();
```
marksreeves: To me this still looks like you are creating a component model before checking if it is needed. This is what we want you to avoid.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -110,12 +133,104 @@ protected boolean doHandleRequest(final Request request) {
...
+	/**
+	 * Reset validation state.
+	 */
+	private void resetValidationState() {
+		// if User Model exists it will be returned, othewise Shared Model is returned
+		final FileWidgetModel componentModel = getComponentModel();
+		// If Shared Model is returned then both fileType and fileSize are always valid
+		// If User Model is returned then reset back to true
+		if (!componentModel.validFileSize || !componentModel.validFileType) {
+		    // If the state has changed then we know for certain a user model exists
```
jonathanaustin: getOrCreateComponentModel before update.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
```diff
@@ -0,0 +1,111 @@
...
+	private static final Log LOG = LogFactory.getLog(FileUtil.class);
+
+	/**
+	 * Checks if the file item is one among the supplied file types.
+	 * 
+	 * @param newFile the file to be checked
+	 * @param fileTypes allowed file types
+	 * @return {@code true} if file type is valid
+	 */
+	public static boolean validateFileType(final FileItemWrap newFile, final List<String> fileTypes) {
```
jonathanaustin: Handle null arguments...
If newFile is null, return false
IllegalArgument if fileTypes is null

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
kandula-aswin:
```java
final FileWidgetModel componentModel = getComponentModel();
```
If `componentModel` is a **Shared Model** then both `fileType` and `fileSize` is always *True*, then `if` fails.
If `componentModel` is a **User Model** then both `fileType` and `fileSize` can be modified as desired
Do you want to still create another variable inside `if` as below?
```java
final FileWidgetModel userModel = getOrCreateComponentModel();
userModel.validFileType = true;
userModel.validFileSize = true;
```
I added a new unit test `testValidationMultiUIContext` to test valid and invalid file

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
kandula-aswin: If `fileTypes` in WFileWidget is set to `null` then [WFileWidget.getFileTypes()](https://github.com/kandula-aswin/wcomponents/blob/dc3cfa5bd9b33176afe1d332ae7a739d87476e55/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java#L51) returns an empty list.
So should `FileUtil.validateFileType()` return `IllegalArgumentException` or `True` if fileTypes  argument is `null`?

PR #1387, opened 2017-10-17
Validation errors - fix regressions

marksreeves:
Fixes errors and provides a means to send messages regarding components which are not _really_ diagnosable but used to be able to have messages set against them.


wcomponents-core/src/main/resources/schema/ui/v1/fieldIndicator.xsd
```diff
@@ -3,7 +3,7 @@
 	xmlns:ui="https://github.com/bordertech/wcomponents/namespace/ui/v1.0">
 	<xs:element name="fieldindicator">
```
jonathanaustin: Found the mixed camel case of fieldIndicator.xsd and ui:fileindicator a bit distracting at first, but is OK.

wcomponents-core/src/main/resources/schema/ui/v1/fieldIndicator.xsd
```diff
@@ -13,7 +13,7 @@
 					</xs:annotation>
 				</xs:element>
 			</xs:sequence>
-			<xs:attribute name="id" type="xs:ID" use="required">
```
jonathanaustin: If the ID field is not required, do they need to be uniquely identified?
As these are nested in the "diagnostable" components, would their ID be the parent component plus an identifier (eg seq)?

wcomponents-theme/src/main/xslt/wc.ui.fieldindicator.xsl
```diff
@@ -6,13 +6,22 @@
 		Transform for ui:fieldindicator which is output of WFieldWarningIndicator.
 	-->
 	<xsl:template match="ui:fieldindicator">
```
jonathanaustin: For consistency, should this have been fieldIndicator.xsl? But looking at the other xsl there are different camel cases.

wcomponents-core/src/main/resources/schema/ui/v1/fieldIndicator.xsd
marksreeves: it is an artefact of the old WComponents schema which used camel case element names. Git is file-name case insensitive so it would be annoying to rename them.

wcomponents-core/src/main/resources/schema/ui/v1/fieldIndicator.xsd
marksreeves: The is is used for stand-alone WField[Error|Warning]Indicator and not needed for the diagnostic output of Diagnosables. It is a pain but will make sense when we get rid of the XML.

wcomponents-theme/src/main/xslt/wc.ui.fieldindicator.xsl
marksreeves: See previous comment about git being case insensitive for file names!

marksreeves:
Thank you @jonathanaustin

PR #1396, opened 2017-10-20
selectboxSearch fires change immediately a value is changed

ricksbrown:
In days of yore browsers would fire the change event when a field lost focus.
Firefox still behaves this way, even for dropdowns.
Other browsers now fire change every time a dropdown value is changed.

Since browsers today also have a native typeahead the value of the dropdown will change as the user types, triggering any side effects on change, e.g. as an AJAX trigger. Our custom typeahead will then change the value of the dropdown (but currently not fire the change event) meaning the side effects will not correspond to the current value of the dropdown (until the user moves focus).

### DIFF
--- a/wcomponents-theme/src/main/js/wc/ui/selectboxSearch.js
+++ b/wcomponents-theme/src/main/js/wc/ui/selectboxSearch.js
@@ -26,9 +26,22 @@ define(["wc/string/escapeRe",
 					textTrumpsValue: true,
 					minLenSubstring: 3,
 					minLenVal: 1,
-					debounceDelay: 250
+					debounceDelay: 125  // making this too long can be counter-productive
 				},
-				fireOnchange = false,
+				selectionChanged = debounce(function(element) {
+					// programatically changing the select will not fire change so we gots to do it ourselves
+					/*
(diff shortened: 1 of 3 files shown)


wcomponents-theme/src/main/js/wc/ui/selectboxSearch.js
```diff
@@ -26,9 +26,22 @@ define(["wc/string/escapeRe",
...
+					/*
+					 * Note that we used to fire the change event only when the dropdown lost focus,
+					 * in other words, as per any traditional change event.
+					 * Nowadays this is not consistent with the native behaviour of many browsers (Chrome, IE11) which
+					 * fire change as the selection changes. Only Firefox maintains the traditional behaviour.
+					 * This inconsistency can lead to unexpected behaviour since the native typeahead will be firing changes
+					 * while this typeahead will not. If there are side effects (e.g. an AJAX update is triggered) then the
+					 * dropdown may be displaying a value that is not appropriate for the current state (only resolved when the user
+					 * moves on in the form).
+					 */
```
jonathanaustin: Well documented.

Issue #1078, opened 2017-02-03
WTable action constraints not handling selected rows and pagination


jonathanaustin:
A `WTable` with `row selection`, `dynamic pagination` and a `table action` with `constraints` does not cater for `rows` selected on `other` pages.

For example, if the `action` constraint requires at least one row to be selected and the selected row is not on the current page, the `action` button will display a message stating no rows have been selected.

Jonno1809:
Fixed in PR #1402

PR #1402, opened 2017-10-27
Constraint pagination fix

Jonno1809:
Fix for issue #1078


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTableRenderer.java
```diff
@@ -301,11 +301,13 @@ private void paintTableActions(final WTable table, final WebXmlRenderContext ren
 						int minRows = constraint.getMinSelectedRowCount();
 						int maxRows = constraint.getMaxSelectedRowCount();
 						String message = constraint.getMessage();
+						int selectedOnOther = table.getSelectedRows().size();
```
ricksbrown: This looks like it gets the TOTAL selected row count... Shouldn't it only get the OTHER selected row count (i.e. the count of rows selected on all pages except the current one)?

Assuming this code needs to be changed then clearly this has not been adequately unit tested. I suggest you adopt a test-driven approach here - first write the unit test (which should fail) and then make it work.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTableRenderer.java
```diff
@@ -301,11 +301,13 @@ private void paintTableActions(final WTable table, final WebXmlRenderContext ren
...
 						String type = constraint.isError() ? "error" : "warning";
 
 						xml.appendTagOpen("ui:condition");
 						xml.appendOptionalAttribute("minSelectedRows", minRows > 0, minRows);
 						xml.appendOptionalAttribute("maxSelectedRows", maxRows > 0, maxRows);
+						xml.appendOptionalAttribute("selectedOnOther", selectedOnOther);
```
ricksbrown: What does `appendOptionalAttribute` do? 
Will it ever behave any differently to `appendAttribute` in this scenario? 
This line needs to change, see if you can work out how.

wcomponents-theme/src/main/js/wc/ui/table/action.js
```diff
@@ -39,17 +39,21 @@ define(["wc/dom/event",
...
 					max = condition.max;
 
+					otherSelected = condition.otherSelected;
+
 					if (!(min || max)) { // no condition worth testing.
 						return true;
 					}
 
-					selected = (getFilteredGroup(ROW_CONTAINER.findDescendant(table), {filter: filter})).length;
+					selected = parseInt((getFilteredGroup(ROW_CONTAINER.findDescendant(table), {filter: filter})).length)+
```
ricksbrown: Can you explain your use of parseInt here?

wcomponents-theme/src/main/xslt/wc.ui.table.xsl
```diff
@@ -929,6 +929,8 @@
 		<xsl:value-of select="@minSelectedRows"/>
 		<xsl:text>","max":"</xsl:text>
 		<xsl:value-of select="@maxSelectedRows"/>
+		<xsl:text>","otherSelected":"</xsl:text>
+		<xsl:value-of select="@selectedOnOther"/>
```
ricksbrown: This will create generate something like this: `{ "otherSelected": "12" }`
You have based this on existing examples, like `max` and `min` but actually do we need all those double quotes? What do you think?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTableRenderer.java
marksreeves: Depending on how this attribute pans out the schema may need an update to reflect either that the attribute is always present or to declare the assumed value if it is an optional attribute.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTableRenderer.java
Jonno1809: I chose this way as the value would only update when a request is made back to the server. For example, it would start with 0 initially, I would select say 3 items, click the next page, and then it would update with 3. This seemed to work with adding selections, but I didn't take into account going back to a previous page and removing a selection. We want it to instantly update (eg. a button greyed out to normal again) but realised that the way I did it it would only update on refresh/request.

Looking back on it this is a terrible way to do it, and I clearly need to implement more testing. Going to look into different solutions, may possibly go the route we discussed.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTableRenderer.java
Jonno1809: I was tossing up between appendOptionalAttribute or appendAttribute, and chose appendOptionalAttribute as, from my understanding, it would only add the XML attribute when it is not null, and that it wouldn't show when it is. I figured we only really need to know whether there is items selected on other pages when there is a constraint, and not need to know all the time, therefore making it optional.

I've only just realised now though that a) I forgot to add a flag to show only when there is actually a constraint and b) selectedOnOther is never null (will be an empty set at least, but not null) and so it is always showing anyway, so it might as well be appendAttribute. Depending on what solution I find, this attribute may not even be needed in the end.

wcomponents-theme/src/main/js/wc/ui/table/action.js
Jonno1809: I needed to add the integers together (currently selected + other selected) for the total, and if I just did `selected = (getFilteredGroup(ROW_CONTAINER.findDescendant(table), {filter: filter})).length + other` it would return `NaN`. I then tried to parseInt the entire thing but then they would just concatenate like a string, so I had to parseInt both to actually get them to add together.

wcomponents-theme/src/main/xslt/wc.ui.table.xsl
Jonno1809: Probably not, I just like consistency and assumed that you guys had done it that way for a reason.

wcomponents-theme/src/main/js/wc/ui/table/action.js
marksreeves: Given that the length of an Array is an integer if the object is, in fact, an Array what else could be causing the left side of a `+` operator to cause a concatenation? Hint: `parseInt("77trombones") === 77` whereas `parseInt("trombones77") === NaN`. How would you test both operands to ensure they are both typed as expected?

wcomponents-theme/src/main/xslt/wc.ui.table.xsl
marksreeves: Could this be impacting your addition operation (above)? If you are reading this property directly then you will have a String rather than a Number on at least one side of the `+`.

Jonno1809: Yep, that was it.

Jonno1809:
This is not to be merged until an issue regarding AJAX and registers is fixed (tests won't pass otherwise) but updated this request with my changes anyway for anyone to critque. I will update with an issue number once @marksreeves creates one for it.

marksreeves:
See #1433 as that will have to be fixed before you will be able to successfully merge this. It is what is causing your selenium tests to fail.

Jonno1809:
@marksreeves 
I have:
1. Removed the jar and manifest files, and fixed my .gitignore (whoops)

2. Made `selectedOnOther` a private member of the renderer so that it no longer updates the table model during the render. As it is only needed in the renderer (for now) I've removed the getters and setters for selectedOnOther in WTable. Let me know if this isn't what you wanted.

3. Unfortunately simply removing the quotation marks for `min` and `max` returns a syntax error - I believe the cause of this is from buttons that don't have either a max or min constraint (or neither) as from my understanding (correct me if I'm wrong) the renderer uses `appendOptionalAttribute` with a flag of `min/maxRows > 0`. 
As a result, without a min constraint (that is, `minSelectedRows==0`) would output a JSON object with a missing value. 
E.g.
`conditions[{"min":, "max":"3"...` instead of `conditions[{"min":"", "max":"3"...`
The reason `selectedOnOther` works without quotation marks is because it is currently using `appendAttribute` so that it is always `"otherSelected":0` which is valid.

Should I change it to `appendAttribute` or just not worry about it? I'm sure you would have another solution.

Jonno1809:
Added a check to see if either min or max properties are missing and remove from JSON entirely as discussed with @ricksbrown . 
I'd also like to make it so that `otherSelected` is only added when the table is paginated (as that's the only time it is needed) but can't seem to get it working. I have tried `<xsl:if test="ui:pagination">`, `<xsl:if test="ui:pagination/@mode eq 'dynamic'">` and a few others to no avail.

Does anyone know what the syntax should be? Why does `<xsl:if test="ui:pagination">` work in other places but not here? @marksreeves @ricksbrown

marksreeves:
@Jonno1809 the XST `if` test is using the incorrect XPath selector. The selector in your comment (`"ui:pagination"`) will test for a `ui:pagination` child element of the node currently being processed. As the current node is a `ui:condition` descendant of a `ui:action` then you need the `ui:pagination` child (if any) of the first ancestor `ui:table` (first as tables may be nested). Your selector should be:

``` xslt
<xsl:if test="ancestor::ui:table[1]/ui:pagination">
<!-- do stuff -->
</xsl:if>
```

Issue #1014, opened 2016-12-06
ComponentModels being created unnecessarily


darrenhitchman:
Some wComponent fields use getOrCreateComponentModel rather than getComponentModel.  This means that a deepCopy of the model is created without needing to. 

An example below for AbstractWFieldIndicator shows that a quick check of diags.size() at the start of the method would avoid a new model being created.

//should add if (diags.size() > 0) { ... } 
protected void showIndicatorsForComponent(final List<Diagnostic> diags, final int severity) {
		FieldIndicatorModel model = getOrCreateComponentModel();
		model.diagnostics.clear();
		UIContext uic = UIContextHolder.getCurrent();

		for (int i = 0; i < diags.size(); i++) {
			Diagnostic diagnostic = diags.get(i);

			if (diagnostic.getSeverity() == severity && uic == diagnostic.getContext()
					// To support repeated components.
					&& relatedField == diagnostic.getComponent()) {
				model.diagnostics.add(diagnostic);
			}
		}
	}

Other Components could use getComponentModel then check if it is the sharedModel or different

marksreeves:
@jonathanaustin @darrenhitchman I can re-schedule this if you think it important enough to include in 1.3.

PR #1460, opened 2018-02-20
Unnecessary models

Jonno1809:
A lot of low hanging fruit fixes for  #1014 . It's probably not all of them as there is over 550 instances of it but it's a start.


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WBeanComponent.java
```diff
@@ -169,7 +169,9 @@ public Object getBeanId() {
 	 */
 	@Override
 	public void setBeanProperty(final String propertyName) {
-		getOrCreateComponentModel().setBeanProperty(propertyName);
+		if (!propertyName.equals(getBeanProperty())) {
```
marksreeves: What if the propertyName arg is null?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WImageEditor.java
```diff
@@ -103,7 +105,9 @@ public Dimension getSize() {
 	 * @param size the size of the image editor.
 	 */
 	public void setSize(final Dimension size) {
-		getOrCreateComponentModel().size = size;
+		if (!size.equals(getSize())) {
```
marksreeves: This also needs a null test on the size argument

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WLabel.java
```diff
@@ -120,7 +120,9 @@ public void setText(final String text, final Serializable... args) {
 	 * @param args optional arguments for the message format string.
 	 */
 	public void setHint(final String hint, final Serializable... args) {
-		getOrCreateComponentModel().hint = I18nUtilities.asMessage(hint, args);
+		if (!hint.equals(getHint())) {
```
marksreeves: Needs a null test of hint

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMessageBox.java
```diff
@@ -164,8 +166,10 @@ public Type getType() {
 	 * @param args optional arguments for the message format string.
 	 */
 	public void setTitleText(final String title, final Serializable... args) {
-		MessageModel model = getOrCreateComponentModel();
-		model.title = I18nUtilities.asMessage(title, args);
+		if (getTitleText() == null || !getTitleText().equals(title)) {
```
marksreeves: Is there a need to handle the case where `getTitleText() == null` *and* arg `title` is also `null`? Pointless, maybe, but it is a public member so it is notionally possible that an external call is trying to reset the title (by making it null) _just in case_ it has been set when it has not yet been set. Since the aim of this code is to remove the creation of unnecessary models all cases should be handled to avoid the unnecessary model.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/validation/AbstractWFieldIndicator.java
```diff
@@ -52,17 +52,19 @@ public AbstractWFieldIndicator(final WComponent relatedField) {
 	 * @param severity A Diagnostic severity code. e.g. {@link Diagnostic#ERROR}
 	 */
 	protected void showIndicatorsForComponent(final List<Diagnostic> diags, final int severity) {
-		FieldIndicatorModel model = getOrCreateComponentModel();
-		model.diagnostics.clear();
-		UIContext uic = UIContextHolder.getCurrent();
+		if (diags.size() > 0) {
```
marksreeves: Can you be sure diags is not null at this point? Is null a valid value for the size arg or should you throw an illegal argument exception?
What if there is an existing model with diagnostics - should they be cleared if the new list of Diagnostics is null or empty? The are in the old code but in your adaptation an existing diagnostics list is only cleared if the new list has items.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
```diff
@@ -156,9 +156,10 @@ public void setIdName(final String idName) {
 					+ idName
 					+ " must start with a letter and followed by letters, digits and or underscores.");
 		}
-
-		ComponentModel model = getOrCreateComponentModel();
-		model.setIdName(idName);
+		if (getIdName() == null || !getIdName().equals(idName)) {
```
marksreeves: Do we still want to create a new component mode if `getIdName() == null` *and* the idName arg is also null?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
```diff
@@ -476,7 +488,9 @@ public void setThumbnailSize(final Dimension thumbnailSize) {
 						"Thumbnail size cannot have both height and width set to -1.");
 			}
 		}
-		getOrCreateComponentModel().thumbnailSize = thumbnailSize;
+		if (getThumbnailSize() == null || !thumbnailSize.equals(getThumbnailSize())) {
```
marksreeves: Do we still make a component model if `getThumbnailSize() ==null` **and** the `thumbnailSize` arg is `null`?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
```diff
@@ -855,7 +869,9 @@ private boolean selectionsEqual(final List<?> list1, final List<?> list2) {
 	 * @param fileId the file id that has been uploaded successfully
 	 */
 	private void setFileUploadRequestId(final String fileId) {
-		getOrCreateComponentModel().fileUploadRequestId = fileId;
+		if (getFileUploadRequestId() == null || !getFileUploadRequestId().equals(fileId)) {
```
marksreeves: Same as up there ^ `getFileUploadRequestId() == null` and `fileId` is null - do we still need the component model?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WNumberField.java
```diff
@@ -278,7 +278,9 @@ public void setMinValue(final double minValue) {
 	 * @param minValue the minimum allowable value, or null for no minimum.
 	 */
 	public void setMinValue(final BigDecimal minValue) {
-		getOrCreateComponentModel().minValue = minValue;
+		if (getMinValue() == null || !getMinValue().equals(minValue)) {
```
marksreeves: and again ... both null? do we create a new component model?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WPopup.java
```diff
@@ -104,7 +112,9 @@ public String getUrl() {
 	 * @param url the URL to set.
 	 */
 	public void setUrl(final String url) {
-		getOrCreateComponentModel().url = url;
+		if (!url.equals(getUrl())) {
```
marksreeves: null test

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WPopup.java
```diff
@@ -122,7 +132,9 @@ public String getTargetWindow() {
 	 * @param targetWindow the target window name.
 	 */
 	public void setTargetWindow(final String targetWindow) {
-		getOrCreateComponentModel().targetWindow = targetWindow;
+		if (!targetWindow.equals(getTargetWindow())) {
```
marksreeves: needs a null test

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WVideo.java
```diff
@@ -218,7 +220,9 @@ public String getMediaGroup() {
 	 * @param mediaGroup The media group name.
 	 */
 	public void setMediaGroup(final String mediaGroup) {
-		getOrCreateComponentModel().mediaGroup = mediaGroup;
+		if (getMediaGroup() == null || !getMediaGroup().equals(mediaGroup)) {
```
marksreeves: dopn't need a new component model if getMediaGroup() == null and mediaGroup == null.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WVideo.java
```diff
@@ -308,7 +320,9 @@ public String getAltText() {
 	 * @param altText the text to set.
 	 */
 	public void setAltText(final String altText) {
-		getOrCreateComponentModel().altText = altText;
+		if (getAltText() == null || !getAltText().equals(altText)) {
```
marksreeves: if both are null (again - see above) don't need a new component model....

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WWindow.java
```diff
@@ -125,7 +125,9 @@ public String getTitle() {
 	 * @param title The title to set.
 	 */
 	public void setTitle(final String title) {
-		getOrCreateComponentModel().title = title;
+		if (!title.equals(getTitle())) {
```
marksreeves: needs a null test on title

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WBeanComponent.java
Jonno1809: Was going to change it to 
`if (propertyName != null && (getBeanProperty() == null || !propertyName.equals(getBeanProperty())))`
but that wouldn't do anything if the propertyName is null. Can/should a bean property ever be set null?

marksreeves: I think your first snippet is testing the wrong thing. You want something like (as per your other null tests):

``` java
if ( propertyName != null ...
```

The only safe assumption for any public or protected member is that object arguments may be null and the return value of a function which returns anything other than a primitive type may also be null, so yes the return value of `getBeanProperty()` may be null _unless_ `getBeanProperty()` is final **and** the method is guaranteed to not return `null` (for example if it gets a property off a bean and if it is null then returns a new instance of the original return type, but this is **very** unlikely).

marksreeves:
@Jonno1809 I suggest you close this PR, fix the issues and make a new PR into `georgie`

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WBeanComponent.java
Jonno1809: I'm not actually using either code snippet above anymore (see below for when both strings are null). For primitive types I'm using `if (arg != getArg())` which is easy as it compares whether it is null or not.

As for Strings, I've been using `!arg.equals(...)` as they're objects and using the `==` operator checks for the same memory reference which is never likely to be the case here, we only want the value.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMessageBox.java
Jonno1809: To address the problem of both arg and getArg being null when they are Strings (my reason for not using != above), I have used the following to all your other comments regarding creating a model when both are null:
```
if ((title != null && getTitleText() == null)
	|| (title == null && getTitleText() != null)
	|| (getTitleText() != null && !getTitleText().equals(title))) {
        getOrCreateComponentModel.title = title (or whatever)
```

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/validation/AbstractWFieldIndicator.java
Jonno1809: How about this at the beginning?
```
if (diags == null || diags.isEmpty()) {
	throw new IllegalArgumentException("Diagnostic list " + diags + "cannot be null or empty!");
}
```

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractWComponent.java
Jonno1809: In this particular instance an empty or null `idName` throws an IllegalArgumentException so it will never be null.

Jonno1809:
Closing to make a new PR for `georgie`, will link back to this so that we know what comments we're addressing in the new PR.

PR #1473, opened 2018-03-06
Unnecessary models

Jonno1809:
Partial fix for #1014, see PR #1460 for more details.


Jonno1809:
Yeah I like that style better, I sometimes forget when calculating boolean expressions that even though it's "simpler" algebraically doesn't mean it's shorter.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WBeanComponent.java
```diff
@@ -169,7 +169,12 @@ public Object getBeanId() {
 	 */
 	@Override
 	public void setBeanProperty(final String propertyName) {
-		getOrCreateComponentModel().setBeanProperty(propertyName);
+		String currBeanProperty = getBeanProperty();
+
+		if ((propertyName != null && currBeanProperty == null)
+			|| (currBeanProperty != null && (propertyName == null || !currBeanProperty.equals(propertyName)))) {
```
jonathanaustin: This style of `if statement` could be simplified by using `!Objects.equals(currBeanProperty, propertyName)`. Replace this pattern that is used in other areas. It would be good to unit test each of these scenarios to make sure a new model is created or not created.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/validation/AbstractWFieldIndicator.java
```diff
@@ -52,19 +52,26 @@ public AbstractWFieldIndicator(final WComponent relatedField) {
 	 * @param severity A Diagnostic severity code. e.g. {@link Diagnostic#ERROR}
 	 */
 	protected void showIndicatorsForComponent(final List<Diagnostic> diags, final int severity) {
-		FieldIndicatorModel model = getOrCreateComponentModel();
-		model.diagnostics.clear();
-		UIContext uic = UIContextHolder.getCurrent();
+		FieldIndicatorModel model = getComponentModel();
+		if (model != null) {
+			model.diagnostics.clear();
```
jonathanaustin: This `model.diagnostic.clear()` could potentially update the `default` model as it has been retrieved with `getComponentModel` instead of `getOrCreateComponentModel`. As per your other changes, determine if the list has changed first and then get the model and update it.

Jonno1809:
@jonathanaustin are these unit tests good enough?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WNumberField.java
```diff
@@ -325,8 +325,7 @@ public void setMaxValue(final BigDecimal maxValue) {
 		BigDecimal currMax = getMaxValue();
 
 		if ((maxValue != null && currMax == null)
-			|| (maxValue == null && currMax != null)
-			|| (maxValue != null && !(maxValue.compareTo(currMax) == 0))) {
+			|| currMax != null && (maxValue == null || !(maxValue.compareTo(currMax) == 0))) {
```
jonathanaustin: Can this also use `Objects.equals`?

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/AbstractWComponentTestCase.java
```diff
@@ -147,6 +148,95 @@ protected void assertAccessorsCorrect(final WComponent component, final String m
...
+	 * @param userContextValue the value to be used within the user context
+	 * @param setterArgs array matching the variable argument type
+	 */
+	protected void assertComponentModelUsesDefaultOnSameValue(AbstractWComponent wComponent, String method,
+															  Object userContextValue, Object setterArgs[]) {
+		wComponent.setLocked(false);
+		invokeSetMethod(wComponent, method, userContextValue, setterArgs);
+		ComponentModel shared = wComponent.getDefaultModel();
+		wComponent.setLocked(true);
+
```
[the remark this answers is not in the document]
Jonno1809: Yes - just a quick test that `getOrCreateComponentModel` is indeed getting the default model instead of creating a new one (without an active context).

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/AbstractWComponentTestCase.java
```diff
@@ -147,6 +148,95 @@ protected void assertAccessorsCorrect(final WComponent component, final String m
...
+	 * @param method the method used to change the model
+	 * @param userContextValue the value to be used within the user context
+	 * @param setterArgs array matching the variable argument type
+	 */
+	protected void assertNoDuplicateComponentModels(AbstractWComponent wComponent, String method,
+			Object userContextValue, Object[] setterArgs) {
+		wComponent.setLocked(true);
+
+		setActiveContext(createUIContext());
+
```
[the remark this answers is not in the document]
Jonno1809: In this method I'm not aiming to compare against the default model, I'm trying to test whether the model changes after setting the value. The original model before setting a value can often be different to the default model, so comparing it to the default isn't going to help here I don't think. 

I'm not entirely sure how to go about testing it though. Say you have the default model (e.g. `shared = wcomponent.getDefaultModel()`), and then you change a value to be different. This should create a new model (e.g. `model = getComponentModel()`), and so `model != shared`. 

However, say you then set the same value - we need to make sure a new model has not been created. We can call and store `newModel = getComponentModel()` and try `assertSame(model, newModel)` or `assertEquals` but that doesn't really help either as they both point to the same reference and will always be true, regardless of whether a new model was created or not after `invokeSetMethod`.

Holding a shared model won't help, and using getComponentModel() with assertEquals won't help. Does any of this make sense? How should we go about this?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WLabel.java
```diff
@@ -110,7 +111,11 @@ public String getText() {
 	 * @param args optional arguments for the message format string.
 	 */
 	public void setText(final String text, final Serializable... args) {
-		getOrCreateComponentModel().text = I18nUtilities.asMessage(text, args);
+		String currText = getText();
+
```
jonathanaustin: This would probably work 90% of the time, but the value that should be tested against the current value is the response from 'asMessage(text,args)`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WLabel.java
```diff
@@ -120,7 +125,11 @@ public void setText(final String text, final Serializable... args) {
 	 * @param args optional arguments for the message format string.
 	 */
 	public void setHint(final String hint, final Serializable... args) {
-		getOrCreateComponentModel().hint = I18nUtilities.asMessage(hint, args);
+		String currHint = getHint();
+
```
jonathanaustin: This should also test the response from `asMessage` against the current value.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMessageBox.java
```diff
@@ -164,8 +167,12 @@ public Type getType() {
 	 * @param args optional arguments for the message format string.
 	 */
 	public void setTitleText(final String title, final Serializable... args) {
-		MessageModel model = getOrCreateComponentModel();
-		model.title = I18nUtilities.asMessage(title, args);
+		String currTitle = getTitleText();
```
jonathanaustin: Same as before, this should test the response from`asMessage` against the current value.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WNumberField.java
```diff
@@ -280,7 +280,12 @@ public void setMinValue(final double minValue) {
 	 * @param minValue the minimum allowable value, or null for no minimum.
 	 */
 	public void setMinValue(final BigDecimal minValue) {
-		getOrCreateComponentModel().minValue = minValue;
+		BigDecimal currMin = getMinValue();
+
```
jonathanaustin: To be consistent, this should also use the `Objects.equals`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WNumberField.java
```diff
@@ -317,7 +322,12 @@ public void setMaxValue(final double maxValue) {
 	 * @param maxValue the maximum allowable value, or null for no maximum.
 	 */
 	public void setMaxValue(final BigDecimal maxValue) {
-		getOrCreateComponentModel().maxValue = maxValue;
+		BigDecimal currMax = getMaxValue();
+
```
jonathanaustin: This should also use `Objects.equals`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WLabel.java
Jonno1809: Just to clarify - you mean something like this?
```
String currText = getText();
Serializable textToBeSet = I18nUtilites.asMessage(text, args);
if (!Objects.equals(textToBeSet, currText) {
    getOrCreateComponentModel().text = textToBeSet;
}
```

jonathanaustin: This is actually a bit tricky (or maybe a better term is messy).

The utility method `asMessage` can return either a `null`, `String` or `Message`.

`getText` always returns a `String`. If the `model` contains a `Message`, then `getText` transforms it into a `String`.

So to actually check if the `value` is being changed on the `model`, you cannot use `getText` but you have to get the value directly from the model:-

```
Serializable currText = getComponentModel().text;
Serializable textToBeSet = I18nUtilites.asMessage(text, args);
if (!Objects.equals(textToBeSet, currText)) {
    getOrCreateComponentModel().text = textToBeSet;
}
```

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/WMessageBox_Test.java
```diff
@@ -147,4 +148,12 @@ public void testTitleTextAccessors() {
...
 		Assert.assertNull("Default accessible text incorrect", comp.getTitleText());
 	}
+
+	@Test
+	public void testDuplicateComponentModels() {
+		WMessageBox messageBox = new WMessageBox(WMessageBox.Type.SUCCESS);
+		assertNoDuplicateComponentModels(messageBox, "type", WMessageBox.Type.INFO);
+//		No such method, invokeSetMethod doesn't like setting strings
+//		assertNoDuplicateComponentModels(messageBox, "titleText", "Test");
+	}
```
Jonno1809: @jonathanaustin  Just to drag this PR out even longer (lol) do you have any idea why invokeSetMethod won't work when setting Strings? Error message is Caused by: java.lang.NoSuchMethodException: Property 'titleText' has no setter method in class 'class com.github.bordertech.wcomponents.WMessageBox'. This happens on all the others with strings which is why I commented them out.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WLabel.java
```diff
@@ -111,10 +111,11 @@ public String getText() {
 	 * @param args optional arguments for the message format string.
 	 */
 	public void setText(final String text, final Serializable... args) {
-		String currText = getText();
+		Serializable currText = getComponentModel().text;
+		Serializable textToBeSet = I18nUtilities.asMessage(text, args);
 
 		if (!Objects.equals(text, currText)) {
```
jonathanaustin: Changes are looking good, except the equals test needs to be `textToBeSet` and `currText`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMessageBox.java
```diff
@@ -167,11 +167,12 @@ public Type getType() {
 	 * @param args optional arguments for the message format string.
 	 */
 	public void setTitleText(final String title, final Serializable... args) {
-		String currTitle = getTitleText();
+		Serializable currTitle = getComponentModel().title;
+		Serializable titleToBeSet = I18nUtilities.asMessage(title, args);
 
 		if (!Objects.equals(title, currTitle)) {
```
jonathanaustin: This should also be testing `titleToBeSet` and `currTitle`.

wcomponents-core/src/test/java/com/github/bordertech/wcomponents/WMultiFileWidget_Test.java
```diff
@@ -309,6 +309,18 @@ public void testMaxFileSizeAccessors() {
...
+	@Test
+	public void testDuplicateComponentModels() {
+		WMultiFileWidget multiFileWidget = new WMultiFileWidget();
+		assertNoDuplicateComponentModels(multiFileWidget,"maxFileSize", 2012312312);
+		assertNoDuplicateComponentModels(multiFileWidget, "maxFiles", 123);
+		assertNoDuplicateComponentModels(multiFileWidget, "newUpload", true);
+		assertNoDuplicateComponentModels(multiFileWidget, "useThumbnails", true);
+		assertNoDuplicateComponentModels(multiFileWidget, "thumbnailPosition", WLink.ImagePosition.SOUTH);
+		assertNoDuplicateComponentModels(multiFileWidget, "thumbnailSize", new Dimension(22,33));
+//		assertNoDuplicateComponentModels(multiFileWidget, "fileUploadRequestId", "testId"); // No such method exception as it's a private method
```
Jonno1809: @jonathanaustin Setter methods that have an access modifier other than public can't be found by `invokeSetMethod`. When making these methods `public `and uncommenting lines such as this one the tests pass. Is that good enough, or would you rather me write special tests that replicate all the tests in `assertNoDuplicateModels`, place them in this test file, and make the methods `protected`/no modifier?

jonathanaustin: @Jonno1809 At the moment the test class can only access public methods. To fix this is beyond the scope of this PR (even though it needs fixing). Raise a new Issue about the ability for the test harness to access "non-public" methods.
So...for this PR leave it commented out.

Issue #1471, opened 2018-03-05
Mandatory indicator is placed too far from label


Likhasinh:
Mandatory field is placed too far from the field label (aligns to the end of the hint text) when field layout is stacked (as a default) and hint text is 'long-ish'.

marksreeves:
This is now an implementation issue. The core theme has minimal styling around the position of the mandatory indicator which will resolve the root cause of this issue - resolved in #1475.

PR #1475, opened 2018-03-16
Theme strip style

marksreeves:
The purpose of this significant theme change is to move towards WComponents being more easily styled but making the structure more consistent and rigid. This will be needed as we move towards making the themes into modules in npm or similar.

A secondary effect of this is to greatly simplify creation and maintenance of the WComponents theme and implementations of this theme by removing the ant magic from the Sass build cycle and using regular Sass mechanisms. This will help us move the theme build out of Ant.

We are removing much of the complexity of the Sass set up and an enormous number of overridable variables, making implementing a custom theme easier and producing a smaller CSS output - hopefully with much less wasted CSS per implementation.

The structure is also moving towards the ability to include CSS loading per module under http2 so we should get even less wasted style.


marksreeves:
resolves #1471 by removing the in-built styles used to position the mandatory indicators and only applying a minimum style to structure the indicators correctly for each label:input combination (label before or above input and label after checkbox.radio).

PR #1500, opened 2018-04-03
Package update

Jonno1809:
Updated package.json to latest working versions (the version that passes all tests, not necessarily the latest published version) and removed tildes to prevent unwanted updates. 

These updated modules will have to be added to Nexus 3 before merging so that we can use it on the network at work.


pom.xml
```diff
@@ -14,6 +14,7 @@
 	<artifactId>wcomponents-parent</artifactId>
 	<version>1.5.3-SNAPSHOT</version>
 
+
```
marksreeves: This change is not warranted

wcomponents-examples/src/main/java/com/github/bordertech/wcomponents/examples/table/SimplePaginationTableWithSelectablesExample.java
```diff
@@ -1,6 +1,5 @@
 package com.github.bordertech.wcomponents.examples.table;
 
-
```
marksreeves: This change is not warranted

wcomponents-examples/src/test/java/com/github/bordertech/wcomponents/examples/SimplePaginationTableWithSelectablesExample_Test.java
```diff
@@ -160,6 +160,7 @@ public void testActionConstraintsAcrossPagesWithoutRefresh() {
 			driver.findElement(byWComponentPath(SELECT_BUTTON)).isEnabled());
 
 		wTableWebElement = getTable();
+
```
marksreeves: This change is not warranted

wcomponents-theme/src/main/js/wc/ui/table/action.js
```diff
@@ -58,6 +58,7 @@ define(["wc/dom/event",
 					totalSelected = currentSelected + otherSelected;
 
 					if ((min && totalSelected < min) || (max && totalSelected > max)) {
+
```
marksreeves: This change is not warranted

marksreeves:
@Jonno1809 I suggest closing this, throwing away those 46 commits and trying again from a clean base. The actual work to update the npm versions is all we need. The rest just produces unnecessary noise.

Jonno1809:
@marksreeves No idea why or how they got there...will try again

PR #1501, opened 2018-04-03
Update package.json

Jonno1809:
PRv2, now with less unwanted commits!

Updated package.json to latest working versions (the version that passes all tests, not necessarily the latest published version) and removed tildes to prevent unwanted updates.


Jonno1809:
@marksreeves noted

PR #1504, opened 2018-04-03
1502 allow config override to inherit default settings

marksreeves:


marksreeves:
I am going to add some more testing around this and then re-submit.

Issue #1503, opened 2018-04-03
Unused npm modules


marksreeves:
Looks like we may have some npm modules which are no longer required. Could someone check the modules to ensure they are still needed and clean up any which are no longer used.

Maybe start with: 

- **css-b64-images** used to be part of the CSS build a long time ago when we did this stuff but the Sass compiler and compressor stages now does this for us.
- **leadfoot** was added to work around a buggy version of **intern** (IIRC) and so may no longer need to be explicitly required.
- Are **mustache** and **handlebars** both being used? If so we should settle on one of these.
- Is an explicit version og **grunt** still needed?

@Jonno1809 I am tagging you because of your recent work on upgrading these dependencies, let me or @ricksbrown know if you are not able to do this.

PR #1512, opened 2018-05-01
Unused dependencies

Jonno1809:
Removed unused node modules from package.json as per #1503. This includes:
- css-b4-images (unused)
- grunt (appears unused)
- handlebars (moustache is being used instead)
- leadfoot (included in Intern - previous bug that we had issues with was fixed in 1.7.3)
- timezone-js (appears unused, think it was leftover from experimenting with clock.js)
- tracking (unused, was used when @ricksbrown was experimenting with face tracking)

Intern was also updated to latest 3.4.x


wcomponents-theme/build-resource.xml
```diff
@@ -28,10 +28,10 @@
 		</copy>
 
 		<copy todir="${target.rootdir}/resource" overwrite="true">
-			<fileset dir="${common.src.rootdir}/resource" includes="**" excludes="**.xml, **.rdf, **.html, **.svg, **.mustache, **.handlebars" />
```
marksreeves: This will prevent the copy of the theme file `src/main/resource/wc.ui.timeoutWarn.handlebars` but that file has not been updated so requires this resource. The template itself is a simple template so could be a mustache template an could be renamed in the JavaScript.

PR #1577, opened 2018-08-21
Extracted file util functions from imageEdit

kandula-aswin:


wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,101 @@
...
+	FileUtil.prototype.mimeToExt = {
+		"image/jpeg": ["jpeg", "jpg"],
+		"image/png": ["png"],
+		"image/webp": ["webp"]
+	};
+
+	function FileUtil() {
+		/**
+		 * Converts a generic binary blob to a File blob.
+		 * @param {Blob} blob
```
marksreeves: What is blob?

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,101 @@
...
+					filePropertyBag.type = config.type;
+				}
+				name = config.name;
+			}
+			name = name || uid();
+
+
+//			if (typeof File === "function") {
+//				return new File([blob], name, filePropertyBag);
+//			}
```
marksreeves: Clean up this - remove the stuff you don't use and reduce the empty space. This is new: there is no need to copy over the previous thought processes.

ricksbrown:
Regarding the public member `ImageEdit.prototype.mimeToExt`: good catch Mark. 

In this case I think we can remove it - what Aswin has done should be fine and is probably the right way to go in this case.

I'm pretty sure it was public not for reuse, but rather to allow OVERRIDING (from application code). I seem to recall an instance where we had only mapped `jpeg` to `image/jpeg`. This was a PITA and so when I fixed it I guess I made it easier to hack up "at runtime". 

So in this case it should move to `file\utils` and in the unlikely event that something breaks we can take a good look at WHY people are changing the mappings and maybe incorporate into the core code.

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,97 @@
...
+	FileUtil.prototype.mimeToExt = {
+		"image/jpeg": ["jpeg", "jpg"],
+		"image/png": ["png"],
+		"image/webp": ["webp"]
+	};
+
+	function FileUtil() {
+		/**
+		 * Converts a generic binary blob to a File blob.
+		 * @param {Blob} The binary blob.
```
ricksbrown: Where is the parameter name?
Here you have defined a parameter called `The` which has the type `{Blob}`.

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,97 @@
...
+			checkFileExtension(blob);
+			return blob;
+		};
+		
+
+		/**
+		 * Ensures that the file name ends with an extension that matches its mime type.
+		 * @param file The file to check.
+		 */
+		function checkFileExtension(file) {
```
ricksbrown: Do you think this has some generic utility? Should we make this public so it can be reused by other modules?

wcomponents-theme/src/main/js/wc/file/util.js
marksreeves: Good catch - replaced on JSDoc error with another. JSDoc is *not* JavaDoc.

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,71 @@
...
+			},
+			testBlobToFile: function () {
+				var expectedMime = "text/plain",
+					expectedFile = "test.file",
+					blob = new Blob(["This is my blob content"], {type: expectedMime}),
+					file = fileUtil.blobToFile(blob, {
+						name: expectedFile
+					});
+				assert.strictEqual(file.name, expectedFile);
+				assert.strictEqual(file.type, expectedMime);
```
ricksbrown: Perhaps check all the expected file properties https://developer.mozilla.org/en-US/docs/Web/API/File#Properties
(i.e. check that it is really the "duck" we are expecting)

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,71 @@
...
+				var expectedText = "hello world",
+					expectedMime = "text/plain",
+					dataurl = 'data:'+ expectedMime +';base64,' + btoa(expectedText),
+					fileReader = new FileReader(),
+					blob;
+
+				blob = fileUtil.dataURItoBlob(dataurl);
+				new Promise(function (win) {
+					fileReader.onload = function () {
+						assert.strictEqual(fileReader.result, expectedText);
```
ricksbrown: Just for your information (you don't need to change this) I would tend to write my event handlers in a more portable and reusable way (not relying on the scope), i.e.
```javascript
fileReader.onload = function (event) {
    assert.strictEqual(event.target.result, expectedText);
```

In this particular case it really doesn't matter because you can't easily move the event handler AND it's very small

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,126 @@
+define(["wc/dom/uid", "wc/file/getMimeType", "wc/config"], function (uid, getMimeType, wcconfig) {
+	"use strict";
+
+	var instance = new FileUtil();
+
+		/**
+		 * Used when checking the newly created file is named with the correct extension.
+		 * If it does not already match any in the array then the extension at index zero will be appended.
+		 **/
+	FileUtil.prototype.mimeToExt = {
```
ricksbrown: This doesn't (shouldn't?) be public anymore. It was originally public to allow overriding, but this is now achieved via config. Therefore it does not need to be public.

Leaving it public now introduces a new problem, the user may access this property directly, expecting it to reflect the mime map set in the config, but instead they will get unpredictable results:
* If they happen to have previously called `fixFileExtension` they will get the map with config overrides
* If they haven't previously called it they will get the default map

If it MUST be public (it probably doesn't?) then perhaps use an accessor method or similar technique.

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,126 @@
...
+
+		/**
+		 * Ensures that the file name ends with an extension that matches its mime type.
+		 * If the file name does not match the mime type then the appropriate extension will be appended.
+		 * If there are multiple possible extensions the the first will be used.
+		 * @param {Blob} file The file to check.
+		 */
+		this.fixFileExtension = function (file) {
+			var expectedExtension,
+				info = getMimeType({
```
ricksbrown: I know I wrote this but `info` is a terrible variable name that conveys no meaning.
Don't need to change it, naming things is hard, "just sayin'" is all

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,126 @@
...
+					files: [file]
+				});
+			if (info && info.length) {
+				info = info[0];
+
+				// get any configuration overrides.
+				var myMimeTypes = wcconfig.get("wc/file/myMimeTypes");
```
ricksbrown: Should be manually hoisted AND it's very odd that eslint didn't find this and get angry.

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,71 @@
...
+			},
+			testFixFileExtensionMultipleExt: function () {
+				var expectMime = "text/plain",
+					file = new Blob(["Text file content"], {type: expectMime}),
+					expectedExt = fileUtil.mimeToExt[expectMime][0],
+					actualExt;
+				file.lastModifiedDate = new Date();
+				file.name = "testfile.jpg";
+				fileUtil.fixFileExtension(file);
+				assert.strictEqual(file.type, expectMime);
```
ricksbrown: doesn't really belong here?

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,71 @@
...
+				actualExt = file.name.split(".");
+				assert.strictEqual(actualExt[2], expectedExt);
+			},
+			testFixFileExtensionCustomMime: function () {
```
ricksbrown: I think you are relying on the order of test execution (which is "Bad") 
Please move this test to the top of the suite and see what I mean.

I suggest you will need something like setup and teardown (or just try/finally) to:
1. Fetch and store the "default" mime map (the handy dandy new getter will be perfect for this)
2. Run the test
3. Restore the default mime map.

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,71 @@
...
+				var expectMime = "application/json",
+					blob = {hello: "world"},
+					expectedFile = "jsonfile",
+					expectedExt = "json",
+					file;
+				// Register custom mimetypes, this will override default mimetypes
```
ricksbrown: Perhaps we need a different test to assert that the mime map is indeed in the state we expect after the override?

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,71 @@
...
+				wcconfig.set({
+					"application/json": ["json"]
+				}, "wc/file/myMimeTypes");
+				file = new Blob([JSON.stringify(blob)], {type: expectMime});
+				file.lastModifiedDate = new Date();
```
ricksbrown: Does this also need `lastModified`? Probably not, just asking...

wcomponents-theme/src/main/js/wc/file/util.js
kandula-aswin: `vars-on-top` [rule](https://eslint.org/docs/rules/vars-on-top) is not defined in [build-import.xml](https://github.com/BorderTech/wcomponents/blob/3e8e1d947be645e0747274e80b80e2be608fdcfd/wcomponents-theme/build-import.xml#L85).
Adding it now introduces 140 problems.

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,135 @@
+define(["wc/dom/uid", "wc/file/getMimeType", "wc/config"], function (uid, getMimeType, wcconfig) {
+	"use strict";
+
+	var instance = new FileUtil(),
+
+		/**
+		 * Used when checking the newly created file is named with the correct extension.
+		 * If it does not already match any in the array then the extension at index zero will be appended.
```
ricksbrown: This second line of this comment block does not make any sense in this context.

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,135 @@
...
+			"text/richtext": ["rtx"],
+			"text/html": ["htm", "html", "stm"],
+			"text/xml": ["xml"],
+			"application/msword": ["doc", "dot"],
+			"application/vnd.ms-powerpoint": ["pot,", "pps", "ppt"],
+			"application/rtf": ["rtf"],
+			"application/pdf": ["pdf"],
+			"application/vnd.ms-excel": ["xla", "xlc", "xlm", "xls", "xlt", "xlw"]
+		};
+
```
ricksbrown: At the very least this should have JSDoc `@constructor`

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,135 @@
...
+			} else {
+				blob.name = name;
+			}
+			return blob;
+		};
+
+		/**
+		 * Ensures that the file name ends with an extension that matches its mime type.
+		 * If the file name does not match the mime type then the appropriate extension will be appended.
+		 * If there are multiple possible extensions the the first will be used.
```
ricksbrown: typo "the the"

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,135 @@
...
+					if (expectedExtension.indexOf(metadata.ext) < 0) {
+						file.name += "." + expectedExtension[0];
+					}
+				}
+
+			}
+		};
+		
+		/**
+		 * Getter for mimetype to extention map.
```
ricksbrown: typo: extention

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,135 @@
...
+		 * It returns default initailsed map, if not overridden by 'wc/config'.
```
ricksbrown: typo: initailsed

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,135 @@
...
+		mimeToExt = {
+			"image/jpeg": ["jpeg", "jpg"],
+			"image/bmp": ["bmp"],
+			"image/png": ["png"],
+			"image/gif": ["gif"],
+			"image/x-icon": ["ico"],
+			"image/webp": ["webp"],
+			"image/svg+xml": ["svg"],
+			"image/tiff": ["tif", "tiff"],
+			"text/plain": ["bas", "c", "h", "txt", "log", "bat"],
```
ricksbrown: wouldn't "txt" be the most logical first entry for "text/plain"?

kandula-aswin: Agree, interestingly even [mimemap.json](https://github.com/BorderTech/wcomponents/blob/3e8e1d947be645e0747274e80b80e2be608fdcfd/wcomponents-theme/src/main/resource/mimemap.json#L10) has `bas` as first entry.

wcomponents-theme/src/main/js/wc/file/util.js
kandula-aswin: Accept, but shame no jsdoc spell-checker in NetBeans

wcomponents-theme/src/main/js/wc/file/util.js
kandula-aswin: 👍 added little jsdoc too

wcomponents-theme/src/main/js/wc/file/util.js
```diff
@@ -0,0 +1,147 @@
...
+			"text/plain": ["txt", "bas", "c", "h", "log", "bat"],
+			"text/tab-separated-values": ["tsv"],
+			"text/richtext": ["rtx"],
+			"text/html": ["htm", "html", "stm"],
+			"text/xml": ["xml"],
+			"application/msword": ["doc", "dot"],
+			"application/vnd.ms-powerpoint": ["pot,", "pps", "ppt"],
+			"application/rtf": ["rtf"],
+			"application/pdf": ["pdf"],
+			"application/vnd.ms-excel": ["xla", "xlc", "xlm", "xls", "xlt", "xlw"]
```
ricksbrown: What extension would you generally expect an Excel file to be given? Would you expect "xla"?

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,98 @@
...
+				assert.strictEqual(file.type, expectedMime);
+				assert.strictEqual(file.size, content.length);
+				// https://developer.mozilla.org/en-US/docs/Web/API/File#Properties
+				assert.isDefined(file.lastModified, "Check this is a file (not a blob).");
+				assert.strictEqual(file.lastModified, file.lastModifiedDate.getTime());
+				return new Promise(function (win) {
+					var fileReader = new FileReader();
+					fileReader.onload = function () {
+						assert.strictEqual(fileReader.result, content);
+						assert.notStrictEqual(fileReader.result, "content");
```
ricksbrown: If we already have done a strict equal check with `"This is my blob content"` does this second assertion test anything?

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,98 @@
...
+						win();
```
ricksbrown: I would tend to put this in a `finally` block but it's probably ok without it - it's hard to see how the simple assertions above could fail. If they DID fail then the test runner would hang as `win` would never be called. Like I say, just food for thought, probably not necessary here.

wcomponents-theme/src/test/intern/wc.file.util.test.js
kandula-aswin: I have added `notStrictEqual` based on our previous conversation that it is better to test success and failure to make sure the filereader has loaded the correct content (text)

wcomponents-theme/src/test/intern/wc.file.util.test.js
kandula-aswin: you mean `onload` function, to look like this?
```javascript
try {
	assert.strictEqual(fileReader.result, content);
	assert.notStrictEqual(fileReader.result, "content");
} finally {
	win();
}
```

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,98 @@
...
+			},
+			testDataURItoBlob: function () {
+				var expectedText = "hello world",
+					expectedMime = "text/plain",
+					dataurl = 'data:' + expectedMime + ';base64,' + btoa(expectedText),
+					blob;
+
+				blob = fileUtil.dataURItoBlob(dataurl);
+				assert.strictEqual(blob.type, expectedMime);
+				assert.strictEqual(blob.size, expectedText.length);
```
ricksbrown: Hmmm interesting one this one. I'm pretty sure it's "wrong" however it will probably work forever.
The size of the blob is in bytes. Your test assumes that each character corresponds to one byte and since your test string is in the ASCII range it _probably_ always will be (I'm honestly not sure).

It certainly would break if you put a character in there outside the ASCII range. You can change the string to:

`expectedText = "hello world\u00DC"`

In trying this you will discover that `btoa` is broken, so you would need to add this [b64EncodeUnicode](https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#The_Unicode_Problem) function to the test and call it instead of btoa `dataurl = 'data:' + expectedMime + ';base64,' + b64EncodeUnicode(expectedText),`

Then you will see what I mean, this assertion will fail. Perhaps you should simply assert that the size is greater that the length of the expected text...?

kandula-aswin: wow, great find, yes agree

wcomponents-theme/src/test/intern/wc.file.util.test.js
```diff
@@ -0,0 +1,98 @@
...
+					expectedExt = fileUtil.getMimeToExtMap()[expectMime][0],
+					actualExt;
+				file.name = "testfile.jpg";
+				// verify default exists
+				assert.isDefined(fileUtil.getMimeToExtMap()["text/plain"]);
+				fileUtil.fixFileExtension(file);
+				actualExt = file.name.split(".");
+				assert.strictEqual(actualExt[2], expectedExt);
+			}
+		});
```
ricksbrown: Need more tests:
* that a correctly named file is not modified
* the same as above but with one of the "alternate" extensions (e.g. a text/plain file with extension ".c") (this could be combined with the test above)
* what about a test for the case where the mime type is unknown in the map (e.g. `application/vnd.openxmlformats-officedocument.wordprocessingml.document`)?
* `blobToFile` (and anything else with optional args) needs a test with its optional arg absent.

wcomponents-theme/src/test/intern/wc.file.util.test.js
ricksbrown: yep

wcomponents-theme/src/test/intern/wc.file.util.test.js
ricksbrown: sure

wcomponents-theme/src/main/js/wc/file/util.js
ricksbrown: Continued here #1579

kandula-aswin:
thank you :handshake:

PR #1585, opened 2018-09-07
Perform file validation with file extension first, then file mime type

### LINKED ISSUE #1584: File with no MIME type fails validation
For file [upload widgets](https://github.com/BorderTech/wcomponents/wiki/WMultiFileWidget), if application [sets file type(s)](https://github.com/BorderTech/wcomponents/wiki/WMultiFileWidget#limit-file-types), it is validated on the server.

At server an opensource Apache tool is used to [detect](https://tika.apache.org/1.1/detection.html) the MIME type. Where files that do not have [known MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml), tika performs a magic detection and returns a default MIME type.

For example HPE Records Manager electronic file reference is saved with extension `tr5`. Browsers do not recognise it's MIME type as it does not have it, where as Apache Tika detects and returns MIME type of `application/x-tika-msoffice`. Due to mismatch the validation fails.

kandula-aswin:
resolves #1584


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -77,9 +79,19 @@ public void setFileTypes(final String[] types) {
 	}
 
 	/**
-	 * Set each file type as a valid file mime type to be accepted by the WMultiFileWidget.
+	 * Determines the file types accepted by this widget. Note that duplicates are not allowed and these are not case
+	 * sensitive. A file type may be one of the following:
```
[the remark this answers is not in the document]
marksreeves: These *are* the only MIME types. The fourth item in the list covers that. The list of acceptable content mirrors the HTML spec for the `accepts` attribute.

kandula-aswin:
@jonathanaustin 
* updated javadoc to reflect the new implementation of validating the file types (both MIME and Ext)
* introduced new method getMimeType(), allowing developers to implement their own validation
Please review

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -80,16 +86,27 @@ public void setFileTypes(final String[] types) {
...
 	 * <ul>
-	 * <li>The string audio/* (Indicates that sound files are accepted.)</li>
-	 * <li>The string video/* (Indicates that video files are accepted.)</li>
-	 * <li>The string image/* (Indicates that image files are accepted.)</li>
-	 * <li>A valid MIME type with no parameters (Indicates that files of the specified type are accepted.)</li>
-	 * <li>A string whose first character is a "." (U+002E) character (Indicates that files with the specified file
-	 * extension are accepted).</li>
+	 * <li>MIME type </li>
+	 * <li>Extension</li>
 	 * </ul>
```
marksreeves: Restore the original text of this doclet as it is **more correct** than the text with which you have replaced it. @jonathanaustin was incorrect in his earlier assertion.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -80,16 +86,27 @@ public void setFileTypes(final String[] types) {
...
+	 * <li><i>image/jpeg</i> - indicates that only jpeg image files are accepted.</li>
+	 * </ul>
+	 * Setting mime type is more reliable, as the contents of the file is validated against allowed list.
+	 * <br>
+	 * <u>Extension</u>: is the suffix at the end of the file name, succeeding a <i>"."</i> character. It indicates that files with 
+	 * specified extension is allowed. Some example extensions are: 
+	 * <ul>
+	 * <li><i>.txt</i> - indicates any files with extension <b>txt</b> are accepted.</li>
+	 * <li><i>.jpg</i> - indicates any files with extension <b>jpg</b> are accepted.</li>
+	 * </ul>
```
marksreeves: As part of the restoration of the original text you may want to keep this as further elucidation but it is generally redundant. If you do keep this content then remove all `u` elements as they add no value and reduce usability; replace all `i` elements with `em` elements; and replace all `b` elements with `strong` elements.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
```diff
@@ -358,6 +375,20 @@ public FileItemWrap getValue() {
...
+	 * @return an file's mime type, or null if there was no file uploaded
+	 * @throws IOException if there is an error obtaining the input stream from the uploaded file.
+	 */
+	public String getMimeType() throws IOException {
+		if (getFile() != null) {
+			FileUtil.getFileMimeType(getFile());
+		}
+		return null;
+	}
+
```
marksreeves: This function **always** returns `null`. Is that correct?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
```diff
@@ -159,6 +165,21 @@ public FileWidgetUpload getFile(final String fileId) {
...
+	 * @param fileId the file id
+	 * @return an file's mime type, or null if there was no file uploaded
+	 * @throws IOException if there is an error obtaining the input stream from the uploaded file.
+	 */
+	public String getMimeType(final String fileId) throws IOException {
+		if (getFile(fileId) != null) {
+			FileUtil.getFileMimeType(getFile(fileId).getFile());
+		}
+		return null;
+	}
```
marksreeves: This function **always** returns `null`. It appears to be an error.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
```diff
@@ -208,16 +229,27 @@ public void setFileTypes(final String[] types) {
 
 	/**
 	 * Determines the file types accepted by this widget. Note that duplicates are not allowed and these are not case
-	 * sensitive. A file type may be one of the following:
+	 * sensitive.<br>
+	 * The file type(s) can be either:
 	 * <ul>
-	 * <li>The string audio/* (Indicates that sound files are accepted.)</li>
```
marksreeves: Restore the original text of this list as it is **more correct** than that which replaced it.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
```diff
@@ -208,16 +229,27 @@ public void setFileTypes(final String[] types) {
...
+	 * <li><i>image/jpeg</i> - indicates that only jpeg image files are accepted.</li>
+	 * </ul>
+	 * Setting mime type is more reliable, as the contents of the file is validated against allowed list.
+	 * <br>
+	 * <u>Extension</u>: is the suffix at the end of the file name, succeeding a <i>"."</i> character. It indicates that files with 
+	 * specified extension is allowed. Some example extensions are: 
+	 * <ul>
+	 * <li><i>.txt</i> - indicates any files with extension <b>txt</b> are accepted.</li>
+	 * <li><i>.jpg</i> - indicates any files with extension <b>jpg</b> are accepted.</li>
+	 * </ul>
```
marksreeves: If keeping this extended elucidation then fix the markup as per the comment above.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
```diff
@@ -68,18 +68,17 @@ public static boolean validateFileType(final FileItemWrap newFile, final List<St
 			// If it exists
 			if (split.length == 2) {
 				// Then check if it matches supplied extension(s)
-				extMatch = fileExts.stream().anyMatch(ext -> ext.matches("(.*)" + split[1]));
+				extMatch = fileExts.stream().anyMatch(fileExt -> fileExt.equals("." + split[1]));
 			}
 		}
 		// If extension match is unsucessful, then move to fileMimes list
 		boolean mimeMatch = false;
```
marksreeves: is the initial value strictly necessary?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
```diff
@@ -96,6 +95,21 @@ public static boolean validateFileType(final FileItemWrap newFile, final List<St
...
 
+	/**
+	 * Identify the mime type of a file.
+	 * 
+	 * @param file the File to detect.
+	 * @return mime type as detected by Apache tika, otherwise null.
+	 * @throws IOException if the mime type detection fails.
+	 */
+	public static String getFileMimeType(final File file) throws IOException {
+		if (file != null) {
```
marksreeves: This is OK but given this is a public method don't you think it would be safer to throw an illegal argument exception if we don't have a file? After all, the MIME type of `null` is not `null`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
kandula-aswin: while strictly not necessary, if not initialised to `false`, I need to set to false 
* in `if` statement in `try` block,  
* `catch` block, 
* and add new else block after [line 93](https://github.com/BorderTech/wcomponents/blob/4e6e825546737773489ce40ae9f377734e0e50c4/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java#L93)

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
kandula-aswin: Perhaps we 3 should meet and reach a concences on the solution, I know I had seperate conversations and the opinion differed slightly.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
kandula-aswin: ok

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java
kandula-aswin: No, if the widget has a file, then it's mime will be returned, otherwise null. 
I had added [additional class level javadoc](https://github.com/BorderTech/wcomponents/blob/4e6e825546737773489ce40ae9f377734e0e50c4/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WFileWidget.java#L38) indicating the new API, allowing the client to perform custom mime validations

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
kandula-aswin: No, if the widget has a file, then it's mime will be returned, otherwise null.
I had added [additional class level javadoc](https://github.com/BorderTech/wcomponents/blob/4e6e825546737773489ce40ae9f377734e0e50c4/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java#L36) indicating the new API, allowing the client to perform custom mime validations

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
kandula-aswin: ok

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
kandula-aswin: Because it is dealing with File, I felt leaving `IOException` is appropriate. Both illegal argument exception and `IOException` are runtime exceptions, so the client need not catch.
But I accept to change to illegal argument exception if it makes more sense.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
marksreeves: ok

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
```diff
@@ -74,21 +74,17 @@ public static boolean validateFileType(final FileItemWrap newFile, final List<St
 		// If extension match is unsucessful, then move to fileMimes list
 		boolean mimeMatch = false;
 		if (!extMatch && fileMimes.size() > 0) {
-			try {
```
marksreeves: You could reduce the cyclometric complexity significantly by returning true if `extMatch` is `true` before you even get in here. as the final line returns `extMatch || mimeMatch`.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
```diff
@@ -74,21 +74,17 @@ public static boolean validateFileType(final FileItemWrap newFile, final List<St
...
-				final String mimeType = getFileMimeType(newFile);
-				LOG.debug("File mime-type is: " + mimeType);
-				for (String fileMime : fileMimes) {
-					if (StringUtils.equals(mimeType, fileMime)) {
+			final String mimeType = getFileMimeType(newFile);
+			LOG.debug("File mime-type is: " + mimeType);
+			for (String fileMime : fileMimes) {
+				if (StringUtils.equals(mimeType, fileMime)) {
+					mimeMatch = true;
```
marksreeves: the function returns `true` if `mimeMatch` == `true` so you can significantly reduce cyclometric complexity by returning `true` rather than setting the member. This goes for all other of same type in this nested if group.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WMultiFileWidget.java
marksreeves: Where are the unit tests?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/util/FileUtil.java
```diff
@@ -23,68 +24,108 @@
...
+	 * @param file the File to detect.
+	 * @return mime type as detected by Apache tika, otherwise null.
+	 */
+	public static String getFileMimeType(final File file) {
+		if (file != null) {
+			try {
+				final Tika tika = new Tika();
+				return tika.detect(file.getInputStream());
+			} catch (IOException ex) {
+				LOG.error("Invalid file.");
```
jonathanaustin: Might be good to log more details.

kandula-aswin:
Merging PR after discussing with @marksreeves

PR #1601, opened 2018-11-22
vars on top

ricksbrown:
@marksreeves thoughts on this? I know in some places it is a bit ugly but overall worth it?


marksreeves:
Oops: pressed the wrong button and now I am a reviewer. 

I am a fan of vars-on-top in principle, but I am also a fan of no-use-before-define. Enforcing vars-on-top as error rather than warn would also eliminate one of our most inconsistent coding habits as shown below:

``` javascript
// vars-on-top AND no-use-before-define - uncommon but exists
define(["wc/initialise", ...] function (intialise, ...) {
  var instance;

  function Component() {
    // stuff ...
  }

  instance = new Component();
  initialise.register(instance);
  return instance;
}
```

compared with:

``` javascript
// no vars-on-top AND no-use-before-define - this is quite commone
define(["wc/initialise", ...] function (intialise, ...) {

  function Component() {
    // stuff ...
  }

  var instance = new Component();
  initialise.register(instance);
  return instance;
}
```

and finally:

``` javascript
// vars-on-top AND not no-use-before-define - Rick's favourite!
define(["wc/initialise", ...] function (intialise, ...) {
  var instance = new Component();

  function Component() {
    // stuff ...
  }

  initialise.register(instance);
  return instance;
}
```

So I would like to see vars-on-top at error status combined with no-use-before-define: warn but just vars-on-top: warn is a good start.

ricksbrown:
@marksreeves Thanks for the comments and the approve - and yes, that is my fav :grin:
I'll merge it as-is now but we can tweak the severity later.

Issue #1602, opened 2018-11-22
Allow switching to client side debug resource from browser


ricksbrown:
In these dark days it is sometimes not possible to even install (or configure) fiddler when working on a problem.

That makes it hard to rewrite requests or reponses with debug (unminified) resources. This is a particular pain when all you have is Internet Exploder.

I say it would be simply wizard if we were able to switch debug mode on from the browser - I reckon a cookie is the way to go.

PR #1605, opened 2018-11-23
#1602 - allow client to request debug resources

ricksbrown:
@marksreeves the implementation I settled on here is inspired by:
* your comments above
* the discussion we had
* **the practicalities integrating this into the existing implementation**

I went with a cookie but it could easily be changed, enhanced etc...

I envisage the usage like so:

1. Open dev tools
2. Enable debug mode:

```javascript
// Enable for one day - nice touch on another person's machine in case you forget to clean up
require(["wc/dom/cookie"], function(cookie) {
    cookie.create("wcforcedebug", true, 1);
})
```

```javascript
// Enable indefinitely - useful on a dev machine
require(["wc/dom/cookie"], function(cookie) {
    cookie.create("wcforcedebug", true);
})
```
3. Disable debug mode
```javascript
// Disable
require(["wc/dom/cookie"], function(cookie) {
    cookie.create("wcforcedebug", false);
})
```


jonathanaustin:
LGTM

Issue #1606, opened 2018-12-06
Add a set of Time releated compoents to allow access to the Java 8 Time API


JohnMcGuinness:
In Java 8, JSR 310 introduced a new Time API located in the ```java.time``` package. It is now recommended that when dealing with anything time related that classes in that package be used.

I would like to suggest that ```WDateField``` be deprecated and a new component ```WLocalDateField``` replace it. As the [javadoc ](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html)for ```java.time.LocalDate``` says a ```LocalDate``` represents  _'A date without a time-zone in the ISO-8601 calendar system, such as 2007-12-03.'_. This new component would almost be a drop in replacement for ```WDateField```. I don't think this component should try to do anything more complicated than allow the day month and year of a date to be captured. I would like to see this component explicitly **_not handle_** ```java.util.Date``` or ```java.util.Calendar```. By not handling these effectively deprecated classes, applications would be encouraged to move away from their usage and to use the preferred ```java.time``` classes.

In the scenario where time of day is required, but time zone is not required, I think a ```WLocalDateTimeField``` would be very useful. This would be similar to a ```WLocalDateField```, but the API would work with ```LocalDateTime``` instead of ```LocalDate```. An example of such a component is the [Vaadin Date field](https://demo.vaadin.com/sampler/#ui/data-input/dates/popup-date-field). The way that Vaadin handles inputting date and time would be a good starting point.

A ```WTimeZone``` component would also be very useful.

WDYT @jonathanaustin @ricksbrown @marksreeves

PR #1609, opened 2019-01-01
Add a WLocalDateField component

JohnMcGuinness:
Partly addresses #1606.

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WLocalDateField.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WLocalDateField.java
@@ -1,0 +1,457 @@
+package com.github.bordertech.wcomponents;
+
+import com.github.bordertech.wcomponents.autocomplete.AutocompleteUtil;
+import com.github.bordertech.wcomponents.autocomplete.AutocompleteableDate;
+import com.github.bordertech.wcomponents.autocomplete.type.DateType;
+import com.github.bordertech.wcomponents.util.InternalMessages;
+import com.github.bordertech.wcomponents.util.SystemException;
+import com.github.bordertech.wcomponents.util.Util;
+import com.github.bordertech.wcomponents.validation.Diagnostic;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.List;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * <p>
+ * This component is a date entry component. Individual themes may add features such as automatic text formatting and
+ * date picker widgets. Convenience methods exist to set and get the value as a java.time.LocalDate.
+ * </p>
+ *
+ * @author John McGuinness
+ * @since 1.5.15
+ */
+public class WLocalDateField extends AbstractInput implements AjaxTrigger, AjaxTarget, SubordinateTrigger,
+		SubordinateTarget, AutocompleteableDate {
+
+	/**
+	 * The logger instance for this class.
+	 */
+	private static final Log LOG = LogFactory.getLog(WLocalDateField.class);
+
+	/**
+	 * This date format is used internally to exchange dates between the client and server.
+	 */
+	private static final String INTERNAL_DATE_FORMAT = "yyyy-MM-dd";
+
+	/**
+	 * Creates a WLocalDateField.
+	 */
+	public WLocalDateField() {
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void setData(final Object data) {
+		// This override is necessary to maintain other internal state
+		LocalDateFieldModel model = getOrCreateComponentModel();
+		try {
+			super.setData(convertDate(data));
+			model.text = null;
+			model.validDate = true;
+		} catch (SystemException e) {
+			super.setData(data);
+			model.text = data.toString();
+			model.validDate = false;
+		}
+	}
+
+	/**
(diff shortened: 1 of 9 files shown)


JohnMcGuinness:
Looks like this failed for a reason outside my control:
```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:3.3.2:check (checkDependencies) on project wcomponents-parent: Fatal exception(s) analyzing wcomponents-parent: Unable to continue dependency-check analysis.
[ERROR] 	Unable to download the NVD CVE data.
```

marksreeves:
There is a vulnerability update required before any more builds will pass, see https://github.com/BorderTech/wcomponents/pull/1607

marksreeves:
@JohnMcGuinness Can you supply the following please:

- What is the purpose of this component?
- How does it differ from WDateField?
- Does it produce a UI artefact different from WDateField? If so please supply XSLT updates, if not why do we have it again?
- Provide an example showing all of its exxpected common uses (with and without initial content, read-only, as a subordinate controller and target, as an ajax controller and target (these last few can be shown by updating one of the existing portmanteau examples if that is easier).
- Provide a Selenium WebElement extension to test-lib to enable UI testing of the component.

JohnMcGuinness:
@marksreeves the answers to your questions follow below:

- What is the purpose of this component
_This component allows the Java developer to use ```java.time.LocalDate``` to represent a date instead of the (nearly deprecated) ```java.util.Date``` or ```java.util.Calendar```_
- How does it differ from WDateField
_The only difference is that the Java API uses ```java.time.LocalDate``` instead of ```java.util.Date```. The renderer  ```WLocalDateFieldRenderer``` creates the same output as ```WDateFieldRenderer```_
- Does it produce a UI artefact different from WDateField? If so please supply XSLT updates, if not why do we have it again?
_See my answer to the previous question_
- Provide an example showing all of its exxpected common uses (with and without initial content, read-only, as a subordinate controller and target, as an ajax controller and target (these last few can be shown by updating one of the existing portmanteau examples if that is easier).
_See the example ```WLocalDateFieldExample```_
- Provide a Selenium WebElement extension to test-lib to enable UI testing of the component.
_As discussed offline I will do this as soon as I have time_

marksreeves:
@jonathanaustin PTAL ASAP

jonathanaustin:
Making use of `LocalDate` is a good idea. We do need a component that makes it easy for projects to utilise a `LocalDate` and be able to bean bind to `LocalDate` variables.

We do have the option of creating a new date component, but this does feel a bit unnecessary as the UI widget is not changing and we are essentially trying to cater for an internal java implementation that caters for `LocalDate`.

To keep the solution simple and reduce any API confusion, what if we modify the existing `WDateField` to have the ability to handle `Date` and `LocalDate`. A new `localDate` flag could be added as a constructor variable, which would default to `false`. The bean binding would then check this flag to perform the correct binding logic.

`WDateField` could have new accessors that return and accept `LocalDate`.  `WDateField's` internal component model could also be changed to `LocalDate` and then do the necessary conversion for the existing `Date` accessors. The `Date` accessors and new `isLocalDate()` could be marked as `deprecated` in preparation for `Date` being removed (if that ever happens). `getValue()` would have to continue to return `Date` as changing this to `LocalDate` would be a breaking API change.

Let me know what you think.

JohnMcGuinness:
I would prefer to apply the [KISS ](https://en.wikipedia.org/wiki/KISS_principle) principle.

A component for ```java.util.Date``` and a component for ```java.time.LocalDate```.

Maybe one day one will be deprecated or even removed. It's clear what you're binding to, and you are not relying on component config to apply some complex logic in the component to put the correct type of value into your bean.

As an Application developer I just want a simple API. Don't make me have to second guess myself, "Did i set this ```WDateField``` to use ```Date``` or ```LocalDate```", there is no ambiguity if I have a separate component.

jonathanaustin:
Agree. The `KISS` principle is important.

I suppose it depends on where are we applying the `KISS` principle.

If we have two components (ie `WDateField`, `WLocalDateField`), then the KISS principle applies in that the internal java implementation and API of each component is simple. However, this does mean we are maintaining two implementations of a date widget which includes the duplication of validation, min and max logic. What are the chances of `Date` being deleted from JDK? Probably never. So we will then have two date components forever that only differ in how they do their java binding. As they share the same UI Widget, then any changes to the widget will have to be duplicated in both component renderers. So the `KISS` principle is broken as far as maintaining the components and renderers. Also, there could be potential confusion to what a `local date` widget is from the UI widget library point of view. A UI person might think "what is this local time? the user's local time?".

What if we modify `WDatefield` to handle `LocalDate` and `Date`? The `KISS` principle is broken in that the API is not as clean and the internal implementation is a bit more complex in that it needs to know which implementation to bind (ie a flag). But as `WComponents` is a widget library, we are not creating a new widget just based on the java type used in binding. The `KISS` principle applies in that we have no duplication of code and if the date UI widget is changed, then there is only one component to modify.

Ideally it would be nice to just switch to `LocalDate` but this is not backwards compatible. But if we deprecate any reference to `Date` and if `Date` is ever removed then we can simply remove the deprecated code.

Its a coin toss. There are pros and cons for both.

But.... I have been thinking.... if we did go for two components, rather than call it `WLocalDateField`, lets call it `WDatePicker` which does not imply the java implementation, but is simply a fresh version of `WDateField` using java 8 features. We can deprecate `WDateField` with a comment telling developers to use `WDatePicker` instead.

As `WDateField` cannot extend `WDatePicker` directly (because of `getValue()` return type), to solve the duplicate code and cost of maintaining two components we could create an `AbstractDatePicker` (based on `LocalDate`) that both components extend. This means they could also share the same renderer.

So.....

With a bit more work and creating a new abstract date class, we could go for two components, the deprecated `WDateField` and shiny new `WDatePicker`.

What do you think?

@marksreeves PTAL

PR #1628, opened 2019-07-05
Partial Date Opt-in and other things

ricksbrown:
Jul 05, 2019: All unit tests pass in all browsers and it seems to function however I would like to further simplify XSLT and make some other changes before merging, time permitting.

Other things include:
* Live build of theme resources as you save files (no need to rebuild theme every time)
* Widget knows how to render itself as a DOM element
* Fix broken image capture
* Fix broken image editor
* Fix ajax error handling in `wc/ui/ajax/processResponse`
* Upgrade to Intern 4 (this is a significant change)

Note: this PR is too huge and complex for anyone to review.


ricksbrown:
Note: I think I'll move this to web-components v1

PR #1639, opened 2019-08-14
1638 fix xslt

marksreeves:


marksreeves:
close due to sonar issues - try from push to bt

PR #1643, opened 2019-09-01
Replaced static margin rendering method with a default interface method

JohnMcGuinness:
Discussed with @marksreeves


JohnMcGuinness:
Addresses #1644

ricksbrown:
@JohnMcGuinness If you merge the new commits from upstream your build will no longer fail - at least not for reasons that are beyond your control.

Apologies for the trouble.

JohnMcGuinness:
I would say that a ```WPanelRenderer`` is-a ```MarginRenderer```. Because it renders a margin for a panel.

jonathanaustin:
Maybe the other thing to consider is that `default` methods were mainly introduced in Java 8 to help make Interfaces backward compatible so implementing classes dont break when new "methods" are introduced (not so much for "new" interfaces). We also need to be careful of the "Diamond" multiple inheritance problem using default methods.

I might be missing something. I cant see the benefit of converting to an Interface as each class implementing `RenderMargin` still has to call the `renderMethod` the same as calling the `Util` method. There is not actually any functionality "inherited" as it still needs to be manually invoked.

We have often thought of redesigning the whole Rendering paradigm for components to automatically render based on the interfaces implemented by the component. So in this case a base renderer would automatically handle any interface the component has (eg Marginable, Disableable, AjaxTarget etc....). We had actually thought of pulling more "common" code out of the existing renders into Utility classes as a step towards this.

So maybe keeping the Utility class is better. Any ideas on improving the renderers is welcome.

JohnMcGuinness:
@ricksbrown recommended that I reopen this for further discussion.

wdyt @shreckdavid @jonathanaustin @marksreeves @ricksbrown

JohnMcGuinness:
Actually reopening this time

Issue #1626, opened 2019-05-22
WTabset.setActiveTab should handle WTab as parameter


sgardner0:
I expected WTabSet.setActiveTab(Wcomponent) to work if I pass the WTab to activate. Instead it expects the content bizarrely.

I suggest the WTabSet.setActiveTab(WComponent content) should check if (content instanceof WTab) then get the index of that tab, else do as usual. So it works for both WTab and it's content.

Thanks Sonja

JohnMcGuinness:
@jonathanaustin happy to look at this. Should it just be a method that takes a WTab gets the content and calls the existing method? Or did you have something else in mind?

PR #1656, opened 2019-09-10
Added WTab based equivalent for some methods in WTabSet

JohnMcGuinness:
Addresses #1626. If this is ok then I'll add some tests.


JohnMcGuinness:
@jonathanaustin should I mark the content based methods as deprecated?

ricksbrown:
@JohnMcGuinness for this and any other PRs it would be great if you could update the changelog.

wcomponents-core/pom.xml
```diff
@@ -272,6 +272,24 @@
...
+	  <scope>test</scope>
+	  <version>1.5</version>
+	  <exclusions>
+	   <exclusion>
+	    <groupId>org.seleniumhq.selenium</groupId>
+	    <artifactId>selenium-remote-driver</artifactId>
+	   </exclusion>
+	  </exclusions>
+	 </dependency>
 	</dependencies>
```
jonathanaustin: Not sure why these selenium dependencies have been added (also the indenting is a mix of space and tab)

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTabSet.java
```diff
@@ -496,6 +496,19 @@ public void setActiveTab(final WComponent content) {
...
+	 * @param tab the active tab.
+	 */
+	public void setActiveTab(final WTab tab) {
+		int index = getTabIndex(tab);
+
+		if (index != -1) {
+			setActiveIndex(index);
+		}
+	}
+
```
jonathanaustin: What do you think...to be consistent and avoid code duplication, all the new `WTab` methods could just call the existing `content` methods like you did with `isActive` method.
```
public void setActiveTab(final WTab tab){
    setActiveTab(tab.getContent());
}
```

wcomponents-core/pom.xml
JohnMcGuinness: Weird. I'm not sure why this is here either. I'll remove it.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/WTabSet.java
JohnMcGuinness: Sounds good. I'll update to do this.

JohnMcGuinness:
This closes #1626

Issue #1492, opened 2018-03-27
theme build without Ant


marksreeves:
Two parter:

1. get rid of ant
2. get rid of Maven in favour of a node-based solution.

@ricksbrown

ricksbrown:
#1653 Takes some important steps in this direction. 
It should be a lot easier now.

ricksbrown:
done ages ago

PR #1659, opened 2019-09-13
Port JS build from Ant to Node

ricksbrown:
This is part 1, there is much more to come. So far the JS build and the test build are off Ant.

Still to do: 

* sass build
* xslt build
* sub-implementation build
* developer mode (automatic rebuild as you change code, possibly hot loading in browser)
* massive documentation update to cover all the new tools


ricksbrown:
#1492

PR #1662, opened 2019-09-18
Filing off some of the rough edges so far...

ricksbrown:
* Add reset method to consoleColor
* Stop browsers complaining about resource prefetch
* Make shed.unsubscribe public
* Update lib build to allow for sourcemaps
* Reinstate custom sass-lint rules
* Connect image build to hot module loader
* Update image build to export build method
* Fix linter failOnError flag

And continuing to fix everything... (yes I started that with a conjunction, if you think that's wrong, you're wrong!)


ricksbrown:
Relates to #1492

PR #1670, opened 2019-10-02
Table column footer

### LINKED ISSUE #1645: Allow for tfoot in WTable
Feature request from consuming project.
Allow for table footer in WTable. The original request was for a totals row in the tfoot but this should not be a specific implementation to total columns. Allow a footer with rows as per head and body. Row content to be determined by each application.

jonathanaustin:
WTableColumn now supports a column footer renderer. Closes #1645

When creating a table column a footer renderer can be included in the constructor.

``` java
table.addColumn(new WTableColumn("DOB", new WDateField(), new WText("My Footer")));
```

Rendering column footers is disabled by default, so the footers must be enabled:

``` java
table.setRenderColumnFooters(true);
```


jonathanaustin:
Good pickup. I will adjust the footer rendering to handle the extra columns.

jonathanaustin:
@marksreeves Have changed the renderer to handle the extra columns needed for expand and select controls and included some new unit tests. Have also updated the `WTableOptionsExample` to show/hide footers.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WTableRenderer.java
```diff
@@ -439,6 +439,14 @@ private void paintColumnHeaderFooter(final WTable table, final WebXmlRenderConte
 			// Footers
 			xml.appendTag("tfoot");
 			xml.appendTag("tr");
+			// Allow for expandable column
+			if (table.getExpandMode() != ExpandMode.NONE) {
+				xml.append("<td/>");
+			}
+			// Allow for selectable column
+			if (table.getSelectMode() != SelectMode.NONE) {
+				xml.append("<td/>");
```
marksreeves: So the problem with rendering a HTML tfoot and associated elements is they have to be in the correct form and the td element is not self-closing so should be output as `<td></td>` rather than `<td/>`. It will work with the self-closing element but it isn't right.

jonathanaustin: Agree. Will fix. What a mess self closing tags are HTML vs XHTML.

jonathanaustin:
@marksreeves Have also updated the `table.xsd` to allow for the `tfoot` element. Have used a `xs:any` for the tfoot element. It would be nice to actually refer to a `html:tfoot` element definition from some where but we same to have the same problem in other schemas. Also made all the WTable Renderer tests include schema validation which had never been run before.

jonathanaustin:
@marksreeves thanks mark. For my own curiosity and also happy to change. What would make the schema better?

marksreeves:
> @marksreeves thanks mark. For my own curiosity and also happy to change. What would make the schema better?

The `xs:any` as an element following `ui:tbody` allows any element from another namespace to be added, not just a `tfoot`. The element could have been limited to a `tfoot` by including the relevant HTML schema, but there is not a schema for HTML5 so that only leaves the option of a custom XML element `ui:tfoot` - and no-one wants that.

jonathanaustin:
Having a `html` schema would be awesome. I struggled with the same thing that using the `xs:any` allows for `any` element instead of a `html:tfoot`. I noticed we have other schemas using the `xs:any` so at least we are consistent. I will merge the request.

PR #1678, opened 2019-11-05
Use custom events for shed

JohnMcGuinness:
Updated collapsibleToggle to use custom events for SHED.

@ricksbrown is this okay?

### DIFF
--- a/wcomponents-theme/src/main/js/wc/ui/tabset.js
+++ b/wcomponents-theme/src/main/js/wc/ui/tabset.js
@@ -464,32 +464,33 @@ function(toArray, ariaAnalog, formUpdateManager, getFilteredGroup, initialise, s
 		 * @function module:wc/ui/tabset.shedObserver
 		 * @protected
 		 * @override
-		 * @param {Element} element The element on which the shed action acted.
-		 * @param {String} action The type of shed event. One of EXPAND, COLLAPSE, SELECT or DESELECT.
+		 * @param {Event} $event The shed event that fired.
 		 */
-		this.shedObserver = function (element, action) {
+		this.shedObserver = function ($event) {
+			var element = $event.target,
+				action = $event.type;
 			if (element) {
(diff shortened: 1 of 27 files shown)


wcomponents-theme/src/main/js/wc/ui/tabset.js
```diff
@@ -464,32 +464,33 @@ function(toArray, ariaAnalog, formUpdateManager, getFilteredGroup, initialise, s
...
 							break;
 						default:
 							console.warn("Unknown action", action);
 							break;
 					}
-				} else if ((action === shed.actions.DISABLE || action === shed.actions.ENABLE) && TABLIST.isOneOfMe(element)) {
+				} else if ((action === shed.events.DISABLE || action === shed.events.ENABLE) && TABLIST.isOneOfMe(element)) {
 					// if the tablist is disabled or enabled, diable/enable all the tabs.
 					Array.prototype.forEach.call(this.ITEM.findDescendants(element), function (next) {
 						shed[action](next);
```
JohnMcGuinness: Would the above line ```shed[action](next)``` be OK. Although no tests are failing, I think that would end up being something like ```shed['wcdisable'](next)```. There's no ```wcdisable``` method on shed, so should it be mapped back to ```disable``` or ```enable```.

JohnMcGuinness:
@ricksbrown I've made all of the initial changes that I think need to be made. I'm unsure if there should be new tests, but I'm sure you'll let me know once you review.

wcomponents-theme/src/main/js/wc/ui/tabset.js
JohnMcGuinness: I think ```shed.events[action]``` would map it back.

ricksbrown: Yes this looks like it won't work as is - mapping it back would be an easy solution

PR #1685, opened 2019-11-21
Fixed incorrect logic in shed.isHidden

JohnMcGuinness:


wcomponents-theme/src/test/intern/wc.dom.shed.test.js
```diff
@@ -534,6 +534,36 @@ define(["intern!object", "intern/chai!assert", "intern/resources/test.utils!"],
...
+				assert.isTrue(controller.isHidden(child));
+			},
+			testIsHiddenCommentNodeChildOfElement: function() {
+				var parent = document.createElement("span");
+				var child = parent.appendChild(document.createComment("comment"));
+				assert.isFalse(controller.isHidden(parent));
+				assert.isFalse(controller.isHidden(child));
+				controller.hide(parent);
+				assert.isTrue(controller.isHidden(parent));
+				assert.isFalse(controller.isHidden(child));
```
ricksbrown: Interesting, not really sure what the right thing to do is here. You think comment nodes are essentially not able to be hidden? Makes no sense either way I guess.

ricksbrown:
This seems good - it was clearly flawed before and this makes it better.
What do you think about the comment node situation? It's not a big deal. Might be worth looking at what other libraries do... 

Pretty sure jquery always returns "false" for comment nodes (just like we do). E.g. `$(commentNode).is(":hidden")` always false regardless.

wcomponents-theme/src/test/intern/wc.dom.shed.test.js
JohnMcGuinness: Comment nodes are not elements, they do not have attributes. So they cannot be hidden.

ricksbrown: Yes, remove the test, we'll leave the behavior undefined.
Alternatively simply assert that an exception is not thrown, maybe this `assert.doesNotThrow`

ricksbrown:
LGTM 👍

PR #1691, opened 2019-12-11
Replace anonymous inner classes with lambdas

JohnMcGuinness:
Replaced most usages of inner classes with lambdas. This addresses the
sonar issue "Anonymous inner classes containing only one method should
become lambdas". Lambdas are usually much more concise and easier to
read.

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractInput.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractInput.java
@@ -303,17 +303,13 @@ public abstract class AbstractInput extends WBeanComponent implements Input {
 		if (getActionOnChange() != null) {
 			final ActionEvent event = new ActionEvent(this, getActionCommand(), getActionObject());
 			final boolean isCAT = isCurrentAjaxTrigger();
-			Runnable later = new Runnable() {
-				@Override
-				public void run() {
-					getActionOnChange().execute(event);
-					if (isCAT && UIContextHolder.getCurrent().getFocussed() == null) {
-						setFocussed();
-					}
-				}
-			};
 
-			invokeLater(later);
+			invokeLater(() -> {
+				getActionOnChange().execute(event);
+				if (isCAT && UIContextHolder.getCurrent().getFocussed() == null) {
+					setFocussed();
+				}
+			});
 		} else if (AjaxHelper.isCurrentAjaxTrigger(this) && UIContextHolder.getCurrent().getFocussed() == null) {
 			setFocussed();
 		}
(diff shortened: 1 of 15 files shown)


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/AbstractInput.java
```diff
@@ -303,17 +303,13 @@ protected void doHandleChanged() {
...
-				}
-			};
 
-			invokeLater(later);
+			invokeLater(() -> {
+				getActionOnChange().execute(event);
+				if (isCAT && UIContextHolder.getCurrent().getFocussed() == null) {
+					setFocussed();
+				}
+			});
```
ricksbrown: This makes sense.

A slightly smaller change would have been to leave the variable assignment like so:

```java
Runnable later = () -> {
	getActionOnChange().execute(event);
	if (isCAT && UIContextHolder.getCurrent().getFocussed() == null) {
		setFocussed();
	}
};

invokeLater(later);
```

It's no big deal tho, I don't see any major advantage or disadvantage in this case.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/layout/UIManager.java
```diff
@@ -62,15 +62,7 @@
...
-	private static final Renderer NULL_RENDERER = new Renderer() {
-		/**
-		 * {@inheritDoc}
-		 */
-		@Override
-		public void render(final WComponent component, final RenderContext renderContext) {
-			// NO-OP
-		}
-	};
+	private static final Renderer NULL_RENDERER = (component, renderContext) -> { /* NO-OP */ };
```
ricksbrown: Not sure if we have a preference for explicitly providing param types or not so sure, why not.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/monitor/UicStatsAsHtml.java
```diff
@@ -67,26 +67,9 @@ private static void writeProfileForTree(final PrintWriter writer,
...
-
-					return diff;
-				}
-			}
-		};
-
-		Collections.sort(statList, comparator);
+		Collections.sort(statList,
+            Comparator.comparing(UicStats.Stat::getModelState)
+                .thenComparing(UicStats.Stat::getClassName));
```
ricksbrown: Well it's unit tested so that's good enough for me.

ricksbrown:
Discussed in principle with @jonathanaustin - changes seem pretty straightforward, merging.

Issue #1699, opened 2020-02-16
SAM(Single Abstract Method) interfaces and classes should be annotated with @FunctionalInterface


JohnMcGuinness:
`@FunctionalInterface` is an informational annotation much like `@Override` that indicates that an interface or abstract class is intended to be a functional interface.

Instances of functional interfaces can be created with lambda expressions, method references, or constructor references.

Here are a list of SAM's in WComponents:

`com.github.bordertech.wcomponents.Action`
`com.github.bordertech.wcomponents.BeanProvider`
`com.github.bordertech.wcomponents.Renderer`
`com.github.bordertech.wcomponents.SelectionToggleable`

PR #1700, opened 2020-02-17
Add @FunctionalIterface to appropriate interfaces and classes

JohnMcGuinness:
Adding `@FunctionalInterface` acts as documentation and assists the
compiler.

`@FunctionalInterface` causes compilation errors to be thrown just like
`@Override`. When `@Override` is added to a member and it doesn't override
anything then the compiler throws a compilation error. In the same way,
`@FunctionalInterface` will cause compilation errors if the class or
interface is not a functional interface.
It also acts as documentation for the developer, indicating that an
instance can be created via a lambda expression, a method reference or a
constructor reference.

Addresses #1699 

WDYT @jonathanaustin


jonathanaustin:
Totally agree that a class like `Action` should be a functional interface. And it makes sense to use Lambda expressions with Actions.

`BeanProvider` is a strong contender for a functional interface but it also gets used as an interface on components and not always used in a "functional manner" (if that makes sense).

`Renderer` and `SelectionTogglable` are not really intended as "functions". Yes they have one abstract method but again are not used in a functional manner.

I think putting `@FunctionalInterface` on `Action` absolutely makes sense. The only other thing we might need to aware of is adding the annotation could potentially be a braking API change if projects have extended this interface. Unlikely, but could happen.

Let me know what you think.

JohnMcGuinness:
Maybe another thing to consider is that if applications are compiling with 1.7 or lower then this will possibly/probably break them. They would need to update to 1.8+.

JohnMcGuinness:
I'll only add the interface to `Action` and `BeanProvider`.

marksreeves:
@JohnMcGuinness please add the update to the changelog to this PR before merging as per the review comment from @jonathanaustin. 

@jonathanaustin if you want changes to a PR (including updating the change log) please don't mark the review as approved - request the changes!

PR #1701, opened 2020-02-24
Session exception

### LINKED ISSUE #1627: New exception type for "Invalid session token on request".
At the moment, the "Invalid session token on request" exception is thrown as a SystemException and caught at the top of the Interceptors and reported in the error log with a full stack trace. This can cause a lot of noise in the error logs. Projects may also want to deal with this error state in a custom way which is hard when it is thrown as the generic SystemException.

To help reduce the noise in the logs and allow projects to intercept this error, a new RuntimeException (eg SessionTokenException) should be created. This exception can still be logged as an error but without the stack trace and allow for custom behavior if needed..

jonathanaustin:
Closes #1627 

To make it easier for projects to identify what type of `SystemException` has occurred (eg invalid session token), this change introduces 4 new exceptions that all extend `SystemException`:
- `SessionTokenException` is when an invalid session token was provided on the request or the session has timed out
- `NoContextException` is when an AJAX request was received that expected a User Context to already exist
- `AjaxTriggerException` is when an invalid AJAX trigger ID was provided on the request
- `TargetableIdException` is when an invalid Targetable ID was provided on the request

There are two scenarios when these exceptions can be thrown (1) on a full page submit or (2) on a AJAX/Content request. The `SessionTokenException` can be thrown in either scenario, but it makes sense the `AjaxTriggerException`, `TargetableIdException`, `NoContextException` can only be thrown on an AJAX/Content request.

So if projects want to take a specific action when these exceptions are thrown then they need to decide which scenario they want to customise.

To provide a custom response for a `SessionTokenException` on page submit, a project needs to use a custom implementation of `FatalErrorPageFactory` where they can check for this exception and then render the appropriate response.

To provide a custom response to an AJAX/Content request is a bit different as the `FatalErrorPage` is not used as only an error response code and message is provided instead. On an AJAX/Content request, any exception is wrapped in an `ErrorCodeEscape` exception which is responsible for setting the response code and message. Therefore, if projects want to intercept this behaviour, they will need to override `handleEscape` in the `ContainerHelper` implementation the project is using. The `ErrorCodeEscape` holds the original exception that caused the error so projects can check for the exception they want to customise.

However, projects will usually only want to modify the page submit error page functionality.

These new exceptions have also been changed to only log as warnings to stop unnecessary noise in error logs.


jonathanaustin:
Thanks for approval. Will merge.

There must be bit of a quirk in the coverage stats as some of the lines they are reporting with no coverage (eg in `AjaxErrorInterceptor`) definitely have been tested.

Issue #1726, opened 2020-05-04
Provide convenience instances of Margins where all sides are the same


JohnMcGuinness:
Currently in order to create a margin where all sides are equal we use ```new Margin(Size.SMALL)``` if there is more than one place in an application where a small margin is required we create a new instance of a Margin with small size.

I propose that we create constants for these types of margins such as ```Margin.ZERO```, ```Margin.SMALL``` and ```Margin.MEDUIM```, etc and deprecate, the ```new Margin(Size)``` constructor. This constructor could be made private. It makes no sense recreating identical instances of these Margins all over the place.

PR #1727, opened 2020-05-04
Provide convenience Margin instances

JohnMcGuinness:
Provided a convenience Margin instance for each Size. Deprecated constructor that takes a single Size.

WDYT @jonathanaustin @marksreeves


JohnMcGuinness:
Address #1726

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/Margin.java
```diff
@@ -111,7 +136,9 @@ private Margin(final Size top, final Size right, final Size bottom, final Size l
 	/**
 	 * A margin equal on all sizes.
 	 * @param all the size of the margin to be used on all sides of the component.
+	 * @deprecated Use Margin.ZERO, Margin.SMALL, Margin.MEDIUM, Margin.LARGE or Margin.XL.
 	 */
+	@Deprecated
 	public Margin(final Size all) {
```
jonathanaustin, line 142: I can understand why we might deprecate the constructor and make it private as this would force projects to use the predefined objects and it is stops projects creating their own "all" margins.
However, we are just providing the "all" objects as a convenience and it is already a public constructor, so my first thoughts are we dont really need to deprecate it. Also, if projects want to extend `Margin` for some reason they would need access to the constructor. If we do deprecate it, maybe it is better to say it will change to `protected` in the future. Also, in the unlikely event we add another `size` to the enum we would have to make sure we make the equivalent Margin object. 
My initial reaction is to not deprecate the constructor. Let me know what you think....

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/Margin.java
```diff
@@ -17,6 +17,31 @@
  */
 public class Margin implements Serializable {
 
+	/**
+	 * A margin with a size of Size.ZERO on all sides.
+	 */
+	public static final Margin ZERO = new Margin(Size.ZERO);
```
jonathanaustin, line 23: I like the idea of providing the "all" margin objects.

Should we name them to indicate they are for all? Like `ZERO_ALL` or something like that....

Also, how about some unit tests for the new all objects.

JohnMcGuinness: I think the names are Okay as they are.

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/Margin.java
JohnMcGuinness: Sure. I removed the deprecation from the constructor.

JohnMcGuinness: I will add the tests.

JohnMcGuinness:
I've added the tests. I'm happy to leave the names as they are without having "all" in the name.

jonathanaustin:
Naming variables and especially constants is hard. As names need to be short and concise but also convey their intended use, I still feel the new constant names should indicate `ALL`. Maybe we need a third opinion as getting APIs right is important. Lets call on Rick... @ricksbrown

JohnMcGuinness:
How about this for a compromise. It acheives the same outcome but, I think, with a nicer API than ALL_ZERO, ALL_SMALL etc.

JohnMcGuinness:
I would argue that Margin should be designed to not be subclassed. A Margin is a Margin and that is all it should be.

If I were designing it in modern Java it would be this.

```java
sealed interface Margin permits Same, Mixed {
// ...
}

public class Same extends Margin {
  public Size getSize()  {/*...*/}
}

public class Mixed implements Margin {
  public Size getTop() {/*...*/}
  public Size getBottom() {/*...*/}
  public Size getLeft() {/*...*/}
  public Size getRight() {/*...*/}
}
```
And then Margin could be used like this (eg in the MarginRenderingUtil):
```java

public static void renderMargin(final Marginable component, final WebXmlRenderContext renderContext) {
  final Margin margin = component.getMargin();

  final XmlStringBuilder xml = renderContext.getWriter();

  if(margin instanceof Same s) {
    xml.appendTagOpen("ui:margin");
    xml.appendAttribute("all", s.getSize().toString());
    xml.appendEnd();
  }
  else if (margin instanceof Mixed m) {
    xml.appendTagOpen("ui:margin");
    xml.appendAttribute("north", m.getTop().toString());
    xml.appendAttribute("south", m.getBottom().toString());
    xml.appendAttribute("east", m.getRight().toString());
    xml.appendAttribute("west", m.getLeft().toString());
  }



```

marksreeves:
@JohnMcGuinness - I will leave the code reviews to Jonathan and Rick but please 

- add a changelog entry
- ensure your changes are thoroughly tested and documented
- provide example usage

Issue #1728, opened 2020-05-04
Improve testing of component accessors by providing references to the methods to be tested


JohnMcGuinness:
Currently testing of accessors such as testing the get/setColumns on WTextField is done as such:

```assertAccessorsCorrect(new WTextField(), "columns", 0, 1, 2);```

I would like to propose updating the ```assertAccessorsCorrect``` method, to remove the String parameter which is the name of the property to get the accessors for, and add parameters for the getter and setter directly. It would look like this:

```assertAccessorsCorrect(new WTextField(), WTextField::getColumns, WTextField::setColumns, 0, 1, 2);```

and the method's signature is:
```
protected <C extends WComponent, T> void assertAccessorsCorrect(
	C component,
	java.util.function.Function<C, T> getter,
	java.util.function.BiConsumer<C, T> setter,
	T initValue,
	T defaultValue, final T userContextValue)
```

The benefits of this include: the accessors must exist at compile time and the type of the values passed must be of the correct type, removing a test dependency on Apache PropertyUtils. Also, possible faster test execution due to no usage of reflection being required, as is used by PropertyUtils.

WDYT @jonathanaustin @marksreeves @ricksbrown

PR #1729, opened 2020-05-04
Example of directly testing accessors

JohnMcGuinness:
Here's an example of how the tests would change. If you agree it's worth doing I update all of the other tests. There are quite a few, hence just an example for now.

Addresses #1728


JohnMcGuinness:
I added another function that can cater for setters with a second argument.

Issue #1732, opened 2020-06-06
Remove usage of deprecated APIs from the framework and examples


JohnMcGuinness:
Over a long period of time many Classes, constructors and methods have been deprecated in WComponents. When an API is deprecated it's a sign to users that it should not be used. How can we expect users to not use them if the framework itself and provided examples use them. I have listed below APIs that are used with WComponents, or examples, that should be updated to remove that usage.

- [x] Replace usage of```WBeanContainer``` class with ```WContainer``` #1733
- [x] Replace usage of ```WHeading.TITLE``` and other constant based heading levels with ```HeadingLevel``` values #1734
- [x] Remove usage of deprecated ```int``` based ```FlowLayout``` constructors and replace with ```Size``` based constructors #1735
- [x] Remove usage of deprecated ```int``` based ```ColumnLayout``` constructors and replace with ```Size``` based constructors #1736
- [x] Remove usage of `SelectMode` APIs. Replace with new `SelectionMode` APIs. #1741 
- [x] Remove usage of deprecated ```int``` based ```Margin``` constructors and replace with ```Size``` based constructors #1739
- [x] Remove usage of ```WProgressBar.UnitType``` from examples. #1750
- [ ] Remove usage of ```BorderLayout``` it deprecated and should not be used.
- [ ] Remove ```WMenuItemGroup``` from examples
- [ ] Remove usage of ```WComponent::getTag``` and ```WComponent::setTag``` from ```WDefinitionList``` #1740
- [ ] Remove usage of ```WFileWidget``` from examples and replace with examples of ```WMultiFileWidget``` if they don't already exist.
- [ ] Remove usage of ```WMenu::getSelectedItems``` and ```WMenu::clearSelectedItems``` from examples and replace with ```WMenu::getSelectedMenuItems``` and ```WMenu::clearSelectedMenuItems```.

JohnMcGuinness:
@marksreeves, @jonathanaustin and @ricksbrown are there any other deprecated classes, constructors or methods that can be added to this list? Is there anything that should be removed from the list above?

PR #1733, opened 2020-06-07
Remove usage of WBeanContainer from framework and examples

JohnMcGuinness:
This PR addresses the first point of #1732

### DIFF
--- a/wcomponents-core/src/test/java/com/github/bordertech/wcomponents/WBeanContainer_Test.java
+++ b/wcomponents-core/src/test/java/com/github/bordertech/wcomponents/WBeanContainer_Test.java
@@ -13,7 +13,7 @@ public class WBeanContainer_Test extends AbstractWComponentTestCase {
 
 	@Test
 	public void testChildAccessors() {
-		WBeanContainer container = new WBeanContainer();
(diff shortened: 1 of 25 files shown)


wcomponents-core/src/test/java/com/github/bordertech/wcomponents/WBeanContainer_Test.java
```diff
@@ -13,7 +13,7 @@
 
 	@Test
 	public void testChildAccessors() {
-		WBeanContainer container = new WBeanContainer();
```
jonathanaustin, line 16: This should not really be changed as it is testing WBeanContainer.

Interestingly, there is no WContainer_Test. As this change is playing in this area, it might be worthwhile writing a WContainer_Test as well.

JohnMcGuinness: Yep, makes sense. I reverted the changes in this file back to WBeanContainer.

JohnMcGuinness: I added WContainer_Test.

PR #1734, opened 2020-06-07
Ensure consistent usage of HeadingLevel

JohnMcGuinness:
Replaced usage of ```WHeading.TITLE``` and other constant based heading levels with ```HeadingLevel``` values.

### DIFF
--- a/wcomponents-core/src/test/java/com/github/bordertech/wcomponents/render/webxml/WHeadingRenderer_Test.java
+++ b/wcomponents-core/src/test/java/com/github/bordertech/wcomponents/render/webxml/WHeadingRenderer_Test.java
@@ -1,5 +1,6 @@
 package com.github.bordertech.wcomponents.render.webxml;
 
+import com.github.bordertech.wcomponents.HeadingLevel;
 import com.github.bordertech.wcomponents.Margin;
 import com.github.bordertech.wcomponents.Size;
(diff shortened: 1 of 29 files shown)


wcomponents-core/src/test/java/com/github/bordertech/wcomponents/render/webxml/WHeadingRenderer_Test.java
```diff
@@ -30,27 +31,27 @@ public void testRendererCorrectlyConfigured() {
 	public void testPaint() throws IOException, SAXException, XpathException {
 		final String text = "WHeading_Test.testPaint.heading";
 
-		WHeading heading = new WHeading(WHeading.TITLE, text);
+		WHeading heading = new WHeading(HeadingLevel.H1, text);
```
jonathanaustin, line 34: As these tests are being changed to use HeadingLevel (which they should be), it highlights the unit tests for WHeading are lacking. This would be an opportunity to improve the unit tests for WHeading, particularly around the conversion of the old Heading Level Types to Heading Level.

JohnMcGuinness: Added a test to WHeading_Test to test the conversion.

JohnMcGuinness:
Partly Addresses #1732

PR #1735, opened 2020-06-07
Deprecated FlowLayout constructors must not be used

JohnMcGuinness:
Removed usage of deprecated ```int``` based ```FlowLayout``` constructors and replaced with ```Size``` based constructors.


JohnMcGuinness:
Partly addresses #1732

PR #1750, opened 2020-07-28
Remove usage of progressbar UnitType from examples

JohnMcGuinness:
Removed usage of ```WProgressBar.UnitType``` from examples.


jonathanaustin (approved):
LGTM

PR #1770, opened 2021-02-08
Latest qa and travis configuration

jonathanaustin:
* Update to latest qa-parent.
* Update travis.yml to build WComponents with JDK 11 and Node V10 to support sonarcloud analysis. However, WComponents is still built and published with JDK 8.
* Added a temporary profile in wcomponents-core to exclude test ObjectGraphDump_Test from JDK 9 and above until the reflection code that fails due to the changes in JDK 9 are fixed or removed.


ricksbrown:
Looks like the encoding of the degrees symbol º has changed - I think yours is actually right?

ricksbrown:
LGTM

ricksbrown (approved):
pot file converted to UTF-8 should be fine
focal is definitely the way to go

PR #1772, opened 2021-02-15
Fix access key labels on Buttons

### LINKED ISSUE #1769: Button Access Key labels not showing
Access keys for buttons are working but their labels do not appear.

This seems to be related to an old change when the renderer for buttons was changed (#1465) from outputting XML to HTML and the access keys span was not included.

The access keys need to implemented into the button renderer.

jonathanaustin:
* Reimplement rendering of Access Key labels for buttons.
* Added new interface AccessKeyable and helper util AccessKeyRendererUtil.

Closes #1769.

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/AccessKeyRendererUtil.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/AccessKeyRendererUtil.java
@@ -8,7 +8,7 @@ import com.github.bordertech.wcomponents.util.Util;
 /**
  * Utility methods for rendering access key details.
  */
-public class AccessKeyRendererUtil {
+public final class AccessKeyRendererUtil {
 
 	/**
 	 * Prevent instantiation of utility class.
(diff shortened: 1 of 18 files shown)


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/AccessKeyRendererUtil.java
```diff
@@ -0,0 +1,58 @@
...
+		// Span with Access key label and standard tooltip id suffix
+		xml.appendTagOpen("span");
+		xml.appendAttribute("id", component.getId() + "_wctt");
+		xml.appendAttribute("role", "tooltip");
+		xml.appendAttribute("hidden", "hidden");
+		xml.appendClose();
+		xml.append(key);
+		xml.appendEndTag("span");
+
+	}
```
jonathanaustin, line 56: @marksreeves was just double checking the [access key xsl](https://github.com/BorderTech/wcomponents/blob/wcomponents-parent-1.5.0/wcomponents-theme/src/main/xslt/wc.common.attributes.xsl) and noticed the `aria-describedby` attribute that has the ID of the span holding the label. Could you please confirm this is needed on the button HTML and I will include it when an access key has been set.

marksreeves: Not required. Should not be on the access key XML either. Make the access key pop-up span aria-hidden as access key exposure to users is in the realm of the AT.

marksreeves (approved):

Issue #1782, opened 2021-06-17
Add a convenience method addAll() to MutableContainer


JohnMcGuinness:
Currently when creating components that have multiple children, a developer is required to write code like the following:

```java
public class MyContainer extends WPanel {
  public MyContainer() {
    final WPanel child1 = new WPanel();
    final WPanel child2 = new WPanel();
    super.add(child1);
    super.add(child2);
  }
}
```

To slighlty reduce the amount of code required I would prefer to write this:

```java
public class MyContainer extends WPanel {
  public MyContainer() {
    final WPanel child1 = new WPanel();
    final WPanel child2 = new WPanel();
    super.addAll(child1, child2);
  }
}
```

Aside from the slight reduction of code that the application developer is required to write, it also has the benefit of moving into one place the code where all (or atleast some of) the children are added, making it easier to see the order that children have been added.

PR #1783, opened 2021-06-17
Add a convenience method addAll to MutableContainer

JohnMcGuinness:
Being able to add all, or a group of, children at once will make code easier to read.

Addresses #1782

WDYT @jonathanaustin @ricksbrown @marksreeves

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/MutableContainer.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/MutableContainer.java
@@ -16,6 +16,13 @@ public interface MutableContainer extends NamingContextable {
 	void add(WComponent component);
 
 	/**
+	 * Adds the given components as "shared" children of this component.
+	 *
+	 * @param components the components to add.
+	 */
+	void addAll(WComponent... components);
+
+	/**
 	 * Removes the given component from this components "shared" list of children.
 	 *
 	 * @param aChild the child component to remove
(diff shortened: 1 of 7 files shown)


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/MutableContainer.java
```diff
@@ -15,6 +15,13 @@
 	 */
 	void add(WComponent component);
 
+	/**
+	 * Adds the given components as "shared" children of this component.
+	 *
+	 * @param components the components to add.
+	 */
+	void addAll(WComponent... components);
```
jonathanaustin, line 23: If this can help projects, then I cant see a problem with it. However, to avoid this being a breaking API change (as projects could be implementing their own containers), what if you provided a default implementation on the interface instead.

This could also avoid adding another "package" protected "addAll" method to AbstractWComponent, as we have never liked those methods and they should never of been on AbstractWComponent.

jonathanaustin (changes requested):

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/MutableContainer.java
JohnMcGuinness: I agree in principle with the suggestion. However, removing the method from ```AbstractWComponent``` and adding a default method to the ```MutableContainer``` inteface seems to cause compilation errors. I would have expected that to work.

JohnMcGuinness: Actually the problem is just removing the method from ```AbstractWComponent```. I added the default method to the interface also, that should stop any issue for any projects that are implementing there own ```MutableContainer```s.

Does this address all of your concerns?

JohnMcGuinness: I've implemented the requested changes.

jonathanaustin (approved):
LGTM

PR #1800, opened 2022-06-30
Remove usage of theme's classList module

JohnMcGuinness:
Removed all usage of wc/dom/classList.js. As a first step of removing support for IE.


ricksbrown (approved):
This is great - very nice cleanup in an age where this DOM API is universal, less is definitely more.  This simplifies complexity and maintenance overhead.

Most important of all, it benefits end users 📱 because the size of the download will be reduced, every bit (and byte) counts.

Note: one possible side effect this particular change _could have been_ a new requirement for null checking the element before using the `.classList` property. However this is not necessary because `wc/dom/classList` was as close as possible to native classList and also did not do any null checks. This means we are safe knowing there will be no new null pointer exceptions.

There were a few gotchas in there, this task required care and attention.

Great work ✨and thanks 🙇

You have merge rights - go for it.

Issue #1804, opened 2023-01-09
Replace usage of var in javascript with let or const


JohnMcGuinness:
Modern JavaScript  recommends that let or const should be used instead of var. All browsers now support let and const.

PR #1805, opened 2023-01-09
Replace var usage with let or const in backToTop.js

JohnMcGuinness:
var is no longer recommended. usage should be replaced with let or const.

#1804


ricksbrown (approved):
Looks good thanks, and thanks for fixing the typo.

Issue #1806, opened 2023-01-10
Remove fixes and polyfills from theme where possible.


JohnMcGuinness:
Numerous fixes and polyfills have been added to the theme, usually to remedy some issue with IE. Since IE is no longer supported by MS it is time to start removing the IE fixes. Other non-IE fixes may also be possible to remove as well.

PR #1807, opened 2023-01-10
Remove inlineBlock_ie8 fix

JohnMcGuinness:
IE is no longer supported by MS. It should be safe to remove this knowing that IE8 is no longer supported by MS or WComponents.

#1806


ricksbrown (approved):
Thanks @JohnMcGuinness.

Oh and also...

DIE INTERNET EXPLORER, DIE! 🧨

PR #1809, opened 2023-01-10
Remove ecma5 polyfills

JohnMcGuinness:
Remove the now redundant ES5 polyfills.


ricksbrown:
Thanks.  Yes these are in all modern and even not-so-modern browsers now - https://caniuse.com/es5
Less is definitely more and this is worth culling.

The tests for some of these are still there (e.g. Function.prototype.bind) but that's OK for now, they have no impact on users. We can remove them in an upcoming PR I guess.

PR #1810, opened 2023-01-12
Update files in wcajax to use let and const

JohnMcGuinness:


wcomponents-theme/src/main/js/wc/ajax/Trigger.js
```diff
@@ -191,7 +192,7 @@ function(tag, event, serialize, Widget, getAncestorOrSelf, ajax, formUpdateManag
 	 * @returns {Element} The element which is expected to fire the trigger.
 	 */
 	function getElement(trigger) {
-		var result, element;
+		let result, element;
 		if (trigger.id) {
 			if ((element = document.getElementById(trigger.id))) {
```
ricksbrown, line 197: I would have moved element here, still as a `let` and pulled it out of the if test + assignment, like this:

```javascript
let element = document.getElementById(trigger.id);
if (element) {
```

It's not really worth changing unless you feel like it, and you probably decided just to go minimal.

JohnMcGuinness: I actually made that kind of change in one or two other places. I'll update it.

ricksbrown: I'm sure this is obvious but just to ensure I don't mislead you...

The test + assignment actually never made sense here (at the top level), but it does make some sense in the `else if` blocks below. 

I guess the obvious alternative would be to not do the whole "single return function" thing and have a bunch of returns in there instead of `else if` blocks.

ricksbrown:
@JohnMcGuinness Oops I wrote my last message to you in the PR merge commit message by mistake. I am task-overloaded.

All I was saying is, native modules are there now. Feel free to adopt (or not) as you touch a module for let/const, but manually for now - am not quite confident enough for a scripted replace yet.

PR #1840, opened 2024-03-18
Feature/rileys tests

rm132:


wcomponents-theme/src/test/spec/wc.file.getFileSize.test.mjs
```diff
@@ -13,5 +13,10 @@ describe("wc/file/getFileSize", () => {
 		addFilesToInput(element, [{ value: "SixteenChars.gif", type: "image/gif" }]);
 		expect(getFileSize(element)[0]).toBe(16);
 	});
+	/*it("Throw type error", () => {
```
ricksbrown: What's this one?
I would suggest making a decision, either delete it or reinstate it.

wcomponents-theme/src/test/spec/wc.ui.printButton.test.mjs
```diff
@@ -0,0 +1,28 @@
+import "wc/ui/printButton.mjs";
+import domTesting from "@testing-library/dom";
+
+describe("wc/ui/printButton", () => {
+	const linkId = "ui-printbutton-1",
+		testContent = `<button class="wc-printbutton" data-testid="${linkId}" type="button"  >Click me<button>`;
```
ricksbrown: spot the typo / bug in your HTML

wcomponents-theme/src/test/spec/wc.ui.printButton.test.mjs
```diff
@@ -0,0 +1,28 @@
...
+	let testHolder;
+
+	beforeAll(() => {
+		testHolder = document.body.appendChild(document.createElement("div"));
+		testHolder.innerHTML = testContent;
+	});
+
+	it("test", function() {
```
ricksbrown: Ok fair enough there is only one test but still... Maybe a slightly better name?
What is this test doing, perhaps something like:

it("sets preventDefault on the print button click event",

ricksbrown (approved):
LGTM
Nice work ✨ 

Wax On Wax Off, Paint the Fence, Sand the Floor...

PR #1841, opened 2024-03-20
Port redirect off of XML/XSLT

philipcaisip:
Removes client-side redirects off of the XSLT registration scripts. Uses HTML custom elements, like #1837.

Currently it seems to be functionally correct but the code does define the redirect tag in two different spots ([one in WrongStepAjaxInterceptor](https://github.com/BorderTech/wcomponents/blob/c682a6cf75f9840adcc2eeb9e85824a20874524b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java#L39C1-L42) and [one in ContentRenderer](https://github.com/BorderTech/wcomponents/blob/c682a6cf75f9840adcc2eeb9e85824a20874524b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/render/webxml/WContentRenderer.java#L19-L22)). This likely ought to be in one spot but I wasn't sure which one spot.

### DIFF
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
@@ -175,7 +175,7 @@ public class WrongStepAjaxInterceptor extends InterceptorComponent {
 		writer.print("<ui:ajaxtarget id=\"" + triggerId + "\" action=\"replace\">");
 
 		// Redirect URL
-		writer.print(String.format("<%s url=\"%s\"></%s>", TAG_REDIRECT, redirectUrl, TAG_REDIRECT));
+		writer.print(String.format("<%1$s url=\"%2$s\"></%1$s>", TAG_REDIRECT, redirectUrl));
 
 		writer.print("</ui:ajaxtarget>");
 		writer.print("</ui:ajaxresponse>");
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
@@ -175,7 +175,7 @@ public class WrongStepAjaxInterceptor extends InterceptorComponent {
 		writer.print("<ui:ajaxtarget id=\"" + triggerId + "\" action=\"replace\">");
 
 		// Redirect URL
-		writer.print(String.format("<%1$s url=\"%2$s\"></%1$s>", TAG_REDIRECT, redirectUrl));
+		writer.print(String.format("<%s url=\"%s\"></%1$s>", TAG_REDIRECT, redirectUrl));
 
 		writer.print("</ui:ajaxtarget>");
 		writer.print("</ui:ajaxresponse>");
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
@@ -35,6 +35,10 @@ import org.apache.commons.logging.LogFactory;
  * @author Jonathan Austin
  */
 public class WrongStepAjaxInterceptor extends InterceptorComponent {
+
+	/**
+	 * Tag/name for redirect HTML elements.
+	 */
 	public static final String TAG_REDIRECT = "wc-redirect";
 
 	/**
--- a/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
+++ b/wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
@@ -35,6 +35,7 @@ import org.apache.commons.logging.LogFactory;
  * @author Jonathan Austin
  */
 public class WrongStepAjaxInterceptor extends InterceptorComponent {
+	public static final String TAG_REDIRECT = "wc-redirect";
 
 	/**
 	 * The logger instance for this class.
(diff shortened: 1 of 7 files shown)


wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
```diff
@@ -170,7 +175,7 @@ private void handleRenderRedirect(final PrintWriter writer) {
 		writer.print("<ui:ajaxtarget id=\"" + triggerId + "\" action=\"replace\">");
 
 		// Redirect URL
-		writer.print("<ui:redirect url=\"" + redirectUrl + "\" />");
+		writer.print(String.format("<%s url=\"%s\"></%s>", TAG_REDIRECT, redirectUrl, TAG_REDIRECT));
```
ricksbrown: We don't repeat args to `String.format` and similar interpolation methods. There is always a way to reuse the argument again in a different place. I'll let you work out how 😉

wcomponents-theme/src/test/spec/wc.ui.redirect.test.mjs
```diff
@@ -0,0 +1,81 @@
+import redirect from "wc/ui/redirect.mjs";
+
+describe("wc/ui/redirect", () => {
```
ricksbrown, line 3: You have at least one unused variable. I'll let you find it...
You are also seeing red errors in IntelliJ from TSC... Don't ignore them.

wcomponents-theme/src/test/spec/wc.ui.redirect.test.mjs
```diff
@@ -0,0 +1,81 @@
...
+		return testRegister("https://github.com/BorderTech/wcomponents", false);
+	});
+
+	it("does not redirect to new url given a launch url", () => {
+		spyOn(redirect, "register").and.callThrough();
+		return testRegister("mailto:test@email.com", true);
+	});
+
+	it("recognises attachment links as launch links", () => {
+		spyOn(redirect, "isLaunchUrl").and.callThrough();
```
ricksbrown: I think you may be overusing the spy here (and elsewhere in this suite)? What value does it add?
For example, this test explicitly calls `redirect.isLaunchUrl` and then uses a spy to verify that this call happened.
Could this not be simplified to something like this?

```js
	it("recognises attachment links as launch links", () => {
		expect(redirect.isLaunchUrl("mailto:test@email.com")).toBeTrue();
	});
```

wcomponents-theme/src/test/spec/wc.ui.redirect.test.mjs
```diff
@@ -0,0 +1,81 @@
...
+describe("wc/ui/redirect", () => {
+	// window = Object.create(Window);
+
+	const delay = 50;  // milliseconds to wait for events and stuff to be actioned
+	const origWindow = window.parent;
+	const replaceSpy = jasmine.createSpy();
+
+	beforeEach(() => {
+		window.parent = {
+			location : {
```
ricksbrown: no space before `:` except for the ternary operator

For example this line should be:

`location: {`

wcomponents-theme/src/test/spec/wc.ui.redirect.test.mjs
```diff
@@ -0,0 +1,81 @@
...
+				storedUrl : "",
```
ricksbrown: Regarding `storedUrl` fair solution. I would personally not do it this way mainly because `window.location` actually has a [defined interface](https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-location-interface) and the property `storedUrl` is not part of it right?

Functionally no problem but some of the tools will get angry.
I would simply move `storedUrl` off the object and into the `beforeEach` function as a variable (with `let` not `const`).

philipcaisip:
Hi @ricksbrown, I've made those changes that you mentioned. Regarding the red TSC errors though, I couldn't see any. There are a few yellow warnings but they look unavoidable/minor (just complaining about some weird Jasmine syntax and the mock window). Is this fine?

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
ricksbrown: Ok that works but can you do it while still using `String.format` and without repeating the args?

Clue: read about [argument_index](https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html#syntax)

wcomponents-core/src/main/java/com/github/bordertech/wcomponents/container/WrongStepAjaxInterceptor.java
```diff
@@ -170,7 +175,7 @@ private void handleRenderRedirect(final PrintWriter writer) {
 		writer.print("<ui:ajaxtarget id=\"" + triggerId + "\" action=\"replace\">");
 
 		// Redirect URL
-		writer.print("<ui:redirect url=\"" + redirectUrl + "\" />");
+		writer.print(String.format("<%1$s url=\"%2$s\"></%1$s>", TAG_REDIRECT, redirectUrl));
```
ricksbrown: Nice! 
No need to overuse the argument index, you should be able to only use it where you need it, so it could look something like this:

```java
writer.print(String.format("<%s url=\"%s\"></%1$s>", TAG_REDIRECT, redirectUrl));
```

ricksbrown:
> Hi @ricksbrown, I've made those changes that you mentioned. Regarding the red TSC errors though, I couldn't see any. There are a few yellow warnings but they look unavoidable/minor (just complaining about some weird Jasmine syntax and the mock window). Is this fine?

Yes, it's more a general point. They probably got sorted out when you cleaned up the code a bit.

ricksbrown:
LGTM - you can merge it when you are ready.
EDIT: I'm doing the merge for you, because I want to merge another after that without causing you merge hell.

PR #1842, opened 2024-03-20
Add js unit tests

philipcaisip:
Unit tests for previously uncovered js code.

Found some small bugs in file/validate and date/isLeapYear. This fixes those too.


ricksbrown:
Hi @philipcaisip thanks for this.
See how the SonarCloud Analysis failed? Do you want to have a crack at fixing it?

Also, see how your PR includes some changes to files under `wcomponents-i18n`? What changed, and what made you decide to add those changes to this PR? Are you sure those changes belong here?

ricksbrown:
@philipcaisip before I take a look at this please get the SonarCloud Analysis passing.

philipcaisip:
Hi @ricksbrown, SonarCloud is passing now with no issues. Please take a look.

wcomponents-theme/src/main/js/wc/file/accepted.mjs
```diff
@@ -23,45 +25,64 @@ import getMimeType from "wc/file/getMimeType.mjs";
...
+
+	if (!acceptedType) return true;
+
+	let acceptedTypes = acceptedType.split(",");
+
+	const checkFileBasic = function (mimeType, extension) {
+		return (!mimeType && !extension) ||
+			(extension && acceptedType.indexOf(extension) >= 0) ||
+			(mimeType && acceptedType.indexOf(mimeType) >= 0);
+	}
```
ricksbrown: Missing semicolon

wcomponents-theme/src/test/spec/wc.file.accepted.test.mjs
```diff
@@ -84,4 +84,16 @@ describe("wc/file/accepted", function() {
...
 		expect(accepted(element)).toBeFalse();
 	});
+
+	it("accepts files given as parameter with multiple accepted types", function() {
+		const element = getInput(testHolder, "file6"),
+			fileInfo =  {files: [new File(["foo.gif"], "foo", {type: "image/gif"})]};
+		expect(accepted(element, fileInfo)).toBeTrue();
+	});
+
+	it("does nto accept files given as parameter with mismatched types", function() {
```
ricksbrown: if you are doing another commit may as well fix the typo - nto > not

wcomponents-theme/src/test/spec/wc.file.validate.test.mjs
```diff
@@ -0,0 +1,97 @@
...
+			expect(emptyCallback).toHaveBeenCalledTimes(1);
+			expect(emptyErrback).not.toHaveBeenCalled();
+		} else {
+			expect(emptyCallback).not.toHaveBeenCalled();
+			expect(emptyErrback).toHaveBeenCalledTimes(1);
+		}
+	}
+
+	beforeEach(() => {
+		testHolder = document.body.appendChild(document.createElement("div"));
```
ricksbrown, line 54: It's no big deal, but since you are creating a new `testHolder` for each test, and leaving the previous testHolder lying around, maybe you should have an `afterEach` that removes them too, something like this:

```js
afterEach(() => {
		testHolder.parentElement.removeChild(testHolder);
```

wcomponents-theme/src/main/js/wc/file/accepted.mjs
```diff
@@ -23,45 +25,64 @@ import getMimeType from "wc/file/getMimeType.mjs";
  * @alias module:wc/file/accepted
  * @requires module:wc/file/getMimeType
  * @param {HTMLInputElement} element A "file input" element.
+ * @param fileInfo The array of files to check, inside an object under "files" property
```
ricksbrown: update the JSDoc to reflect that this is an optional parameter `[fileInfo]`

wcomponents-theme/src/main/js/wc/file/accepted.mjs
```diff
@@ -23,45 +25,64 @@ import getMimeType from "wc/file/getMimeType.mjs";
...
+				passed = compareMime(mimeType, next);
+				if (passed) {
+					break;
+				}
+			}
+		}
+		return passed;
+	};
+
+	if (fileInfo) return getMimeType(fileInfo).every(checkFileArray);
```
ricksbrown: While this works fine and there is nothing at all wrong with it, we inherited a style format that insists on curly braces even for single statements. The thinking behind such a rule is that it prevents future bugs, as someone comes along and adds another statement, forgetting to add the curly braces.

I'm not saying I agree with it, but that's how the rest of the codebase is written, so I suggest you add the curly braces to this one too.

wcomponents-theme/src/main/js/wc/file/accepted.mjs
```diff
@@ -23,45 +25,64 @@ import getMimeType from "wc/file/getMimeType.mjs";
...
+		}
+		return false;
+	};
+
+	const checkFileArray = function (/** @type {module:wc/file/getMimeType~fileType} */fileType) {
+		let mimeType = fileType.mime || "",
+			extension = fileType.ext ? (`.${fileType.ext}`) : "",
+			passed = checkFileBasic(mimeType, extension);
+
+		if (passed) {
```
ricksbrown: Do you think this code might be clearer without the `passed` variable? 
Like, we just `return true` instead from all the places, here, line 71 and line 78?
 For example:
```js
if (checkFileBasic(mimeType, extension)) { 
	return true;
}
```
```js
if (extension === next) {
	return true;
}
```

etc

ricksbrown (approved):
Nice work, thanks for this.

PR #1862, opened 2024-06-19
Port rich text field registration off of XSLT

philipcaisip:


JohnMcGuinness:
It should be noted that WebKit does not suppport extending elements

ricksbrown:
> It should be noted that WebKit does not suppport extending elements

Oh darn it! You are right! Nowadays that means Safari.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is#browser_compatibility

philipcaisip:
Hi @ricksbrown, this is ready to review now. RTFs are marked with a custom HTML component that then get passed to TinyMCE, instead of the XSLT doing that. There's also some new unit tests that use a mock TinyMCE, which I've put in a new mock-modules folder.

PR #1867, opened 2025-07-16
Update java dependencies and revert to commons-logging

jonathanaustin:
- Revert to commons-logging instead of jcl-over-slf4j to allow projects to decide on how to handle logging.
- Update project dependencies to the latest versions


ricksbrown:
LGTM 👍

